PR #1, opened 2015-05-28
headers parsing rework 

echampet:
I needed a php lib to parse mail, change some headers and send the result via smtp
To tests zend-mail, i've taken more than 100000 mail from our mail server,
and i've made a small script that:

1) parse it ```$zendmessage1 = Message::fromString($mime);```
2) export it ```$raw1 = $zendmessage1->toString();```
3) parse it again ```$zendmessage2 = Message::fromString($raw1);```
4) export it again ```$raw2 = $zendmessage2->toString();```
5) check if parsing/exporting is stable (```$raw1 === $raw2```)
if not there is a problem
6) try to send it via smtp (postfix configured to forward to /dev/null)
7) compute similar_text between original header and ```$zendmessage1->getHeaders()->toString()```
look at the lowest score to see if some stuff can be improved

This PR is an attempt to fix all bugs that i've encountered playing with these mails

### DIFF
--- a/src/Header/HeaderWrap.php
+++ b/src/Header/HeaderWrap.php
@@ -90,7 +90,7 @@ abstract class HeaderWrap
      */
     public static function mimeEncodeValue($value, $encoding, $lineLength = 998)
     {
-        return Mime::encodeQuotedPrintableHeader($value, $encoding, $lineLength, Headers::EOL);
+        return Mime::encodeBase64Header($value, $encoding, $lineLength, Headers::EOL);
     }
 
     /**
--- a/src/Header/HeaderWrap.php
+++ b/src/Header/HeaderWrap.php
@@ -103,9 +103,7 @@ abstract class HeaderWrap
      */
     public static function mimeDecodeValue($value)
     {
-        $decodedValue = iconv_mime_decode($value, ICONV_MIME_DECODE_CONTINUE_ON_ERROR, 'UTF-8');
-
-        return $decodedValue;
+        return iconv_mime_decode($value, ICONV_MIME_DECODE_CONTINUE_ON_ERROR, 'UTF-8');
     }
 
     /**
--- a/src/Headers.php
+++ b/src/Headers.php
@@ -58,58 +58,49 @@ class Headers implements Countable, Iterator
      * will be lazy loaded)
      *
      * @param  string $string
-     * @param  string $EOL EOL string; defaults to {@link EOL}
+     * @param  string $EOL DEPRECATED
      * @throws Exception\RuntimeException
      * @return Headers
      */
-    public static function fromString($string, $EOL = self::EOL)
+    public static function fromString($string, $EOL = null)
     {
         $headers     = new static();
         $currentLine = '';
         $emptyLine   = 0;
 
         // iterate the header lines, some might be continuations
-        $lines = explode($EOL, $string);
-        $total = count($lines);
-        for ($i = 0; $i < $total; $i += 1) {
-            $line = $lines[$i];
-
+        foreach (preg_split('/\R/', $string) as $line) {
             // Empty line indicates end of headers
             // EXCEPT if there are more lines, in which case, there's a possible error condition
             if (preg_match('/^\s*$/', $line)) {
-                $emptyLine += 1;
+                $emptyLine++;
                 if ($emptyLine > 2) {
-                    throw new Exception\RuntimeException('Malformed header detected');
+                    throw new Exception\RuntimeException('Malformed header detected, too many empty lines');
                 }
                 continue;
             }
-
             if ($emptyLine > 0) {
-                throw new Exception\RuntimeException('Malformed header detected');
+                throw new Exception\RuntimeException('Malformed header detected, too many empty lines');
             }
 
             // check if a header name is present
-            if (preg_match('/^[\x21-\x39\x3B-\x7E]+:.*$/', $line)) {
+            if (preg_match('/^[!-9;-~]+:/', $line)) {
                 if ($currentLine) {
                     // a header name was present, then store the current complete line
                     $headers->addHeaderLine($currentLine);
                 }
                 $currentLine = trim($line);
-                continue;
-            }
-
-            // continuation: append to current line
-            // recover the whitespace that break the line (unfolding, rfc2822#section-2.2.3)
-            if (preg_match('/^\s+.*$/', $line)) {
+            } elseif (preg_match('/^\s+[^\s]/', $line)) {
+                // continuation: append to current line
+                // recover the whitespace that break the line (unfolding, rfc2822#section-2.2.3)
                 $currentLine .= ' ' . trim($line);
-                continue;
+            } else {
+                // Line does not match header format!
+                throw new Exception\RuntimeException(sprintf(
+                    'Line "%s"does not match header format!',
+                    $line
+                ));
             }
-
-            // Line does not match header format!
-            throw new Exception\RuntimeException(sprintf(
-                'Line "%s"does not match header format!',
-                $line
-            ));
         }
         if ($currentLine) {
             $headers->addHeaderLine($currentLine);
--- a/src/Headers.php
+++ b/src/Headers.php
@@ -210,7 +210,7 @@ class Headers implements Countable, Iterator
     {
         if (!is_string($headerFieldNameOrLine)) {
             throw new Exception\InvalidArgumentException(sprintf(
-                '%s expects its first argument to be a string; received "%s"',
+                'addHeaderLine expects its first argument to be a string; received "%s"',
                 (is_object($headerFieldNameOrLine)
                 ? get_class($headerFieldNameOrLine)
                 : gettype($headerFieldNameOrLine))
--- a/src/Headers.php
+++ b/src/Headers.php
@@ -218,13 +218,19 @@ class Headers implements Countable, Iterator
         }
 
         if ($fieldValue === null) {
-            $this->addHeader(Header\GenericHeader::fromString($headerFieldNameOrLine));
+            list($fieldName, $fieldValue) = Header\GenericHeader::splitHeaderLine($headerFieldNameOrLine);
+            $norm = $this->normalizeFieldName($fieldName);
+            $this->headersKeys[] = $norm;
+            $this->headers[] = $headerFieldNameOrLine;
         } elseif (is_array($fieldValue)) {
+            $norm = $this->normalizeFieldName($headerFieldNameOrLine);
             foreach ($fieldValue as $i) {
-                $this->addHeader(Header\GenericMultiHeader::fromString($headerFieldNameOrLine . ':' . $i));
+                $this->headersKeys[] = $norm;
+                $this->headers[] = $headerFieldNameOrLine.": ".$i;
             }
         } else {
-            $this->addHeader(Header\GenericHeader::fromString($headerFieldNameOrLine . ':' . $fieldValue));
+            $this->headersKeys[] = $this->normalizeFieldName($headerFieldNameOrLine);
+            $this->headers[] = $headerFieldNameOrLine.": ".$fieldValue;
         }
 
         return $this;
@@ -300,7 +306,7 @@ class Headers implements Countable, Iterator
         $results = array();
 
         foreach (array_keys($this->headersKeys, $key) as $index) {
-            if ($this->headers[$index] instanceof Header\GenericHeader) {
+            if (is_string($this->headers[$index])) {
                 $results[] = $this->lazyLoadHeader($index);
             } else {
                 $results[] = $this->headers[$index];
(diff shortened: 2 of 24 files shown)


echampet:
fix for the failing unit test is in https://github.com/zendframework/zend-mime/pull/1
(Mime: base64 encoded string must be multiple of 4 chars)
(previously we were checking that zend-mime is broken)

src/Header/HeaderWrap.php
```diff
@@ -90,7 +90,7 @@ protected static function wrapStructuredHeader($value, StructuredInterface $head
      */
     public static function mimeEncodeValue($value, $encoding, $lineLength = 998)
     {
-        return Mime::encodeQuotedPrintableHeader($value, $encoding, $lineLength, Headers::EOL);
+        return Mime::encodeBase64Header($value, $encoding, $lineLength, Headers::EOL);
```
weierophinney: This is a rather big change. Shouldn't users be able to opt-in between quoted-printable and base64?

src/Headers.php
```diff
@@ -58,58 +58,49 @@ class Headers implements Countable, Iterator
...
         $currentLine = '';
         $emptyLine   = 0;
 
         // iterate the header lines, some might be continuations
-        $lines = explode($EOL, $string);
-        $total = count($lines);
-        for ($i = 0; $i < $total; $i += 1) {
-            $line = $lines[$i];
-
+        foreach (preg_split('/\R/', $string) as $line) {
```
weierophinney: WHOA!!! NO!!!!

This does *NOT* follow RFC 2822, and introduces potential for security vectors. Please read [RFC 2822, section 2.1](https://tools.ietf.org/html/rfc2822#section-2.1), which states:

> Messages are divided into lines of characters.  A line is a series of characters that is delimited with the two characters carriage-return and line-feed; that is, the carriage return (CR) character (ASCII value 13) followed immediately by the line feed (LF) character (ASCII value 10).  (The carriage-return/line-feed pair is usually written in this document as "CRLF".)

Additionally, you are undoing a *reported* vulnerability ([ZF2015-04](http://framework.zend.com/security/advisory/ZF2015-04)).

`\R` matches *any* of `\r`, `\n`, or `\r\n`; we specifically were matching only `\r\n` previously to be compliant with the specification and prevent message splitting.

src/Header/HeaderWrap.php
echampet: i agree this is big, but quoted printable encoding has coner case that i don't know how to handle (comma ...)
if it's too big i can put it in another PR

src/Headers.php
```diff
@@ -58,58 +58,49 @@ class Headers implements Countable, Iterator
...
             // Empty line indicates end of headers
             // EXCEPT if there are more lines, in which case, there's a possible error condition
             if (preg_match('/^\s*$/', $line)) {
-                $emptyLine += 1;
+                $emptyLine++;
```
weierophinney: Please revert; the previous is less error-prone. Read "JavaScript: the Good Parts" for an explanation of why. (Yes, the book is on JS; however, many of the principles stated in it apply to any C-style language.)

src/Headers.php
```diff
@@ -58,58 +58,49 @@ class Headers implements Countable, Iterator
...
                 if ($emptyLine > 2) {
-                    throw new Exception\RuntimeException('Malformed header detected');
+                    throw new Exception\RuntimeException('Malformed header detected, too many empty lines');
```
weierophinney: For security reasons, we should never provide more detail than necessary in the exception messages.

src/Headers.php
```diff
@@ -58,58 +58,49 @@ class Headers implements Countable, Iterator
...
             }
-
             if ($emptyLine > 0) {
-                throw new Exception\RuntimeException('Malformed header detected');
+                throw new Exception\RuntimeException('Malformed header detected, too many empty lines');
             }
 
             // check if a header name is present
-            if (preg_match('/^[\x21-\x39\x3B-\x7E]+:.*$/', $line)) {
+            if (preg_match('/^[!-9;-~]+:/', $line)) {
```
weierophinney: The previous regex was better as it detailed the specific ASCII ranges.

src/Headers.php
```diff
@@ -58,58 +58,49 @@ class Headers implements Countable, Iterator
...
                     $headers->addHeaderLine($currentLine);
                 }
                 $currentLine = trim($line);
-                continue;
-            }
-
-            // continuation: append to current line
-            // recover the whitespace that break the line (unfolding, rfc2822#section-2.2.3)
-            if (preg_match('/^\s+.*$/', $line)) {
+            } elseif (preg_match('/^\s+[^\s]/', $line)) {
```
weierophinney: Bad regex; the value on the continuation line *must* be allowed to contain whitespace; your regex prohibits it.

weierophinney: Also, please revert the structure to remove elseif/else blocks, and issue `continue` and `break` statements within the `if` bodies. The original structure promoted readability and reduced complexity.

src/Headers.php
```diff
@@ -227,13 +218,19 @@ public function addHeaderLine($headerFieldNameOrLine, $fieldValue = null)
         }
 
         if ($fieldValue === null) {
-            $this->addHeader(Header\GenericHeader::fromString($headerFieldNameOrLine));
+            list($fieldName, $fieldValue) = Header\GenericHeader::splitHeaderLine($headerFieldNameOrLine);
+            $norm = $this->normalizeFieldName($fieldName);
+            $this->headersKeys[] = $norm;
+            $this->headers[] = $headerFieldNameOrLine;
```
weierophinney: `addHeader()` was previously doing the work you're doing here; why are you duplicating it?

weierophinney: Bigger question: why are you storing the header line and not the header instance?

src/Protocol/Smtp.php
```diff
@@ -273,10 +273,7 @@ public function data($data)
         $this->_send('DATA');
         $this->_expect(354, 120); // Timeout set for 2 minutes as per RFC 2821 4.5.3.2
 
-        // Ensure newlines are CRLF (\r\n)
-        $data = str_replace("\n", "\r\n", str_replace("\r", '', $data));
-
-        foreach (explode(self::EOL, $data) as $line) {
+        foreach (preg_split('/\R/', $data) as $line) {
```
weierophinney: See my note above; this is directly breaking a security fix.

test/Header/GenericHeaderTest.php
```diff
@@ -113,13 +113,13 @@ public function validFieldValuesProvider()
...
 
             // CRLF @group ZF2015-04 cases
-            'newline' => array("xxx yyy\n", '=?UTF-8?Q?xxx=20yyy=0A?=', 'UTF-8'),
-            'cr-lf' => array("xxx yyy\r\n", '=?UTF-8?Q?xxx=20yyy=0D=0A?=', 'UTF-8'),
-            'cr-lf-wsp' => array("xxx yyy\r\n\r\n", '=?UTF-8?Q?xxx=20yyy=0D=0A=0D=0A?=', 'UTF-8'),
-            'multiline' => array("xxx\r\ny\r\nyy", '=?UTF-8?Q?xxx=0D=0Ay=0D=0Ayy?=', 'UTF-8'),
+            'newline' => array("xxx yyy\n", '=?UTF-8?B?eHh4IHl5eQo=?=', 'UTF-8'),
+            'cr-lf' => array("xxx yyy\r\n", '=?UTF-8?B?eHh4IHl5eQ0K?=', 'UTF-8'),
+            'cr-lf-wsp' => array("xxx yyy\r\n\r\n", '=?UTF-8?B?eHh4IHl5eQ0KDQo=?=', 'UTF-8'),
+            'multiline' => array("xxx\r\ny\r\nyy", '=?UTF-8?B?eHh4DQp5DQp5eQ==?=', 'UTF-8'),
```
weierophinney: Again, I have to ask: why are you changing the default encoding mechanism? What does that gain us?

test/HeadersTest.php
```diff
@@ -381,6 +381,7 @@ public function testHeaderCrLfAttackAddHeaderLineWithValue()
         $headers = new Mail\Headers();
         $this->setExpectedException('Zend\Mail\Header\Exception\InvalidArgumentException');
         $headers->addHeaderLine('Fake', "foo-bar\r\n\r\nevilContent");
+        $headers->forceLoading();
```
weierophinney: You're changing a test case, which is an indication of a BC break.

src/Header/HeaderWrap.php
weierophinney: It's a BC break. It needs to be removed at this point, and a really, really strong case needs to be made for changing it.

weierophinney:
Overall, I have zero idea of what you're trying to accomplish in this pull request. It breaks fixes that were introduced for *known* security vectors, and changes the encoding used. Can you detail what you're trying to accomplish please, so we can better evaluate the changes?

src/Protocol/Smtp.php
echampet: in this case no this is not,
with the 2 str_replace
\r => '' (0 eol)
\r\n => \r\n (1 eol)
\n => \r\n (1 eol)
with my preg_split all are 1 eol,
but if my preg_split is unsafe, i'm pretty sure the str_replace is also

Please prove with an exemple that it's insecure :)

src/Header/HeaderWrap.php
echampet: quoted printable is broken, base64 isn't ... but i'll remove it for now :)

src/Headers.php
echampet: a continuation line must contain something that is not tab or space, else it's an empty line !?

echampet: ok for the structure (i prefer mine but i'll not argue)

src/Headers.php
echampet: addHeader is forcing the encoding, which break many mails
(i've parsed more than 100000 mails (only the headers) in my first PR)

for the bigger view, it's to fix the lazyloading, don't need to parse before using the header
(some explanation in the commit also)

test/HeadersTest.php
echampet: headers are now really lazyloaded/lazyparsed, so this now throw exception when you acces them

src/Headers.php
weierophinney: > a continuation line must contain something that is not tab or space, else it's an empty line !?

I misread the intent of the regex; thought I saw a `$` anchoring the end of the line, but there's not. It's fine.

weierophinney: > ok for the structure (i prefer mine but i'll not argue)

I recommend reading up on object calisthenics.

test/HeadersTest.php
weierophinney: So include that information in the pull request summary, as that looks like the important change (or one of the important changes) you're introducing.

echampet:
I'm trying to make zend-mail work with REAL mail :)
i've updated the PR message

weierophinney:
> I'm trying to make zend-mail work with REAL mail :)

That tells me nothing. Tell me what specific issues you're running into, and what your expectations are. We need *explicitness* for issues, not vagueness. The new summary is far more clear, though I'd like to have you delineate the issues you've identified.

echampet:
the original PR is from march (https://github.com/zendframework/zf2/pull/7372),
but nobody looked at it at that time ... (and now i don't remember the specifics)
I'll split this PR in smaller PR, that will help

src/Headers.php
echampet: Mail with mixed line do exists in the wild, and even if it's not rfc compliant, i do want to handle them

for now i fail to find an exemple string that would made us vulnerable
(if you have an exemple in mind ...)

weierophinney: There are several test cases, all referenced as `@group ZF2015-04`, which
must pass after any changes are made. You can likely adapt those to use
just `\n` EOL separators as well to verify.

On Fri, May 29, 2015 at 10:03 AM, echampet <notifications@github.com> wrote:

> In src/Headers.php
> <https://github.com/zendframework/zend-mail/pull/1#discussion_r31332512>:
>
> >      {
> >          $headers     = new static();
> >          $currentLine = '';
> >          $emptyLine   = 0;
> >
> >          // iterate the header lines, some might be continuations
> > -        $lines = explode($EOL, $string);
> > -        $total = count($lines);
> > -        for ($i = 0; $i < $total; $i += 1) {
> > -            $line = $lines[$i];
> > -
> > +        foreach (preg_split('/\R/', $string) as $line) {
>
> Mail with mixed line do exists in the wild, and even if it's not rfc
> compliant, i do want to handle them
>
> for now i fail to find an exemple string that would made us vulnerable
> (if you have an exemple in mind ...)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/zend-mail/pull/1/files#r31332512>.
>



-- 
Matthew Weier O'Phinney
matthew@weierophinney.net
https://mwop.net/

echampet:
i've started to open smaller PR ( #3 #4 #6 #7 ).
@weierophinney thanks for your articles (and review)
i've found a good one about regex: https://swtch.com/~rsc/regexp/regexp1.html

weierophinney:
I'm closing this as the author has begun submitting more granular PRs targeting specific issues.

PR #3, opened 2015-06-01
Headers: fix bad sprintf call

echampet:


src/Headers.php
```diff
@@ -219,7 +219,7 @@ public function addHeaderLine($headerFieldNameOrLine, $fieldValue = null)
     {
         if (!is_string($headerFieldNameOrLine)) {
             throw new Exception\InvalidArgumentException(sprintf(
-                '%s expects its first argument to be a string; received "%s"',
+                'addHeaderLine expects its first argument to be a string; received "%s"',
```
Martin-P: Instead of hardcoding the method name, I think the `__METHOD__` constant is better. That way the class name is also included in the message which makes it far more easier to track the exception.

echampet:
@Martin-P done !

PR #4, opened 2015-06-01
Protocol\Smtp->data(): improve a bit

echampet:
replacing \n with \r\n and then calling explode \r\n is useless
also don't use strpos just to check the first char

### DIFF
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -273,11 +273,12 @@ class Smtp extends AbstractProtocol
         $this->_send('DATA');
         $this->_expect(354, 120); // Timeout set for 2 minutes as per RFC 2821 4.5.3.2
 
-        // Ensure newlines are CRLF (\r\n)
-        $data = str_replace("\n", "\r\n", str_replace("\r", '', $data));
+        // Handle CRLF (\r\n) and LF (\n), remove CR (\r)
+        $data = str_replace("\r", '', $data);
+        $data = explode("\n", $data);
 
-        foreach (explode(self::EOL, $data) as $line) {
-            if (strpos($line, '.') === 0) {
+        foreach ($data as $line) {
+            if (isset($line[0]) && $line[0] === '.') {
                 // Escape lines prefixed with a '.'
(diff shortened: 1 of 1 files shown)


echampet:
see also #5 for a better version

Martin-P:
> replacing \n with \r\n and then calling explode \r\n is useless

Isn't that the same approach as replacing `\r` with an empty string and calling `explode` using `\n`?

> also don't use strpos just to check the first char

Can you show that using `isset($line[0])` *and* `$line[0] === '.'`, where PHP probably casts the string to an array to perform the comparison, is faster than the string function `strpos()`? To my knowledge `strpos()` is about as fast as you can get to determine if a string contains a certain character, but perhaps I'm wrong.

echampet:
Hi @Martin-P,

> Isn't that the same approach as replacing ```\r``` with an empty string and calling explode using ```\n```?

that's what i'm doing, no? (take a look at the diff, there was 2 str_replace, there is now 1, with strictly equivalent result)

strpos will walk all the input to see if there is '.'
the speed difference is small, but for ~660000 lines (my 50M test mail), it's still 30ms
```
$a = str_repeat('a', 100);
echo memory_get_usage().'/'.memory_get_peak_usage()."\n";

$tstart = microtime(true);
for($i=0;$i<660000;$i++) {
    $b = (isset($a[0]) && $a[0] === '.');
}
echo ((microtime(true)-$tstart))."\n";
echo memory_get_usage().'/'.memory_get_peak_usage()."\n";


$a = str_repeat('a', 100);
echo memory_get_usage().'/'.memory_get_peak_usage()."\n";

$tstart = microtime(true);
for($i=0;$i<660000;$i++) {
    $b = (strpos($a, '.') === 0);
}
echo ((microtime(true)-$tstart))."\n";
echo memory_get_usage().'/'.memory_get_peak_usage()."\n";
```
```
227184/251888
0.05359411239624
227640/251888
227640/251888
0.083939075469971
227640/251888
```

echampet:
please also see #6 and #7

PR #5, opened 2015-06-01
Protocol\Smtp->data(): be more memory friendly

echampet:
With huge mails (50M), i'm hitting memory_limit, so i've done some tests
looking at peak memory usage and average time(10 runs)
- str_replace("\r","") + explode("\n") 250M/0,27s (what we use now)
- preg_split("\r?\n") 200M/1,5s
- strtok("\n") + rtrim("\r") 100M/0,27s
- strpos("\n") + substr() 50M/0,50s

All tests were run with 'php -d"memory_limit=256M" -d"opcache.enable_cli=1"'
I've choose strtok because it's the best balance between time and memory

Before this patch, we were removing \r, now we're keeping them,
in any case \r should be encoded

Signed-off-by: Etienne CHAMPETIER <etienne.champetier@fiducial.net>

### DIFF
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -273,16 +273,19 @@ class Smtp extends AbstractProtocol
         $this->_send('DATA');
         $this->_expect(354, 120); // Timeout set for 2 minutes as per RFC 2821 4.5.3.2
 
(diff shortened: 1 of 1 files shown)


echampet:
forget about this one, if you have multiple \n, it's "eating" them

PR #9, opened 2015-06-03
Hotfix zendframework/zf2#7555 - Header\Sender::fromString()

stefanotorresi:
`Header\Sender::fromString()` didn't correctly set the address for some formats (most notably `Sender: email@domain`).

Close zendframework/zf2#7555
Supersede and close #11

### DIFF
--- a/src/Header/Sender.php
+++ b/src/Header/Sender.php
@@ -43,6 +43,12 @@ class Sender implements HeaderInterface
         }
 
         $header     = new static();
+
+        /**
+         * matches the header value so that the email must be enclosed by < > when a name is present
+         * 'name' and 'email' capture groups correspond respectively to 'display-name' and 'addr-spec' in the ABNF
+         * @see https://tools.ietf.org/html/rfc5322#section-3.4
+         */
         $hasMatches = preg_match('/^(?:(?P<name>.+)\s)?(?(name)<|<?)(?P<email>[^\s]+?)(?(name)>|>?)$/', $value, $matches);
 
         if ($hasMatches !== 1) {
--- a/src/Header/Sender.php
+++ b/src/Header/Sender.php
@@ -39,25 +39,23 @@ class Sender implements HeaderInterface
 
         // check to ensure proper header type for this factory
         if (strtolower($name) !== 'sender') {
-            throw new Exception\InvalidArgumentException('Invalid header line for Sender string');
+            throw new Exception\InvalidArgumentException('Invalid header name for Sender string');
         }
 
-        $header      = new static();
-        $senderName  = '';
-        $senderEmail = '';
+        $header     = new static();
+        $hasMatches = preg_match('/^(?:(?P<name>.+)\s)?(?(name)<|<?)(?P<email>[^\s]+?)(?(name)>|>?)$/', $value, $matches);
 
-        // Check for address, and set if found
-        if (preg_match('/^(?P<name>.*?)<(?P<email>[^>]+)>$/', $value, $matches)) {
-            $senderName = trim($matches['name']);
-            if (empty($senderName)) {
-                $senderName = null;
-            }
-            $senderEmail = $matches['email'];
-        } else {
-            $senderEmail = $value;
+        if ($hasMatches !== 1) {
+            throw new Exception\InvalidArgumentException('Invalid header value for Sender string');
         }
 
-        $header->setAddress($senderEmail, $senderName);
+        $senderName = trim($matches['name']);
+
+        if (empty($senderName)) {
+            $senderName = null;
+        }
+
+        $header->setAddress($matches['email'], $senderName);
 
         return $header;
     }
--- a/test/Header/SenderTest.php
+++ b/test/Header/SenderTest.php
@@ -204,4 +204,50 @@ class SenderTest extends \PHPUnit_Framework_TestCase
             'multiline' => ["<foo\r\n@\r\nbar>", $headerInvalidArgumentException, null],
         ];
     }
+
+    /**
+     * @param string $headerString
+     * @param string $expectedName
+     * @param string $expectedEmail
+     * @param string $expectedException
+     *
+     * @dataProvider headerLinesProvider
+     */
+    public function testFromString($headerString, $expectedName, $expectedEmail, $expectedException = null)
+    {
+        if ($expectedException) {
+            $this->setExpectedException($expectedException);
+        }
+
+        $header = Header\Sender::fromString($headerString);
+
+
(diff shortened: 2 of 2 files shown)


test/Header/SenderTest.php
```diff
@@ -190,4 +190,50 @@ public function invalidSenderEncodedDataProvider()
...
+     * @param string $headerString
+     * @param string $expectedName
+     * @param string $expectedEmail
+     * @param string $expectedException
+     *
+     * @dataProvider headerLinesProvider
+     */
+    public function testFromString($headerString, $expectedName, $expectedEmail, $expectedException = null)
+    {
+        if ($expectedException) {
```
Maks3w: Don't add test logic here in the same test case. Instead split the test.

src/Header/Sender.php
```diff
@@ -39,23 +39,23 @@ public static function fromString($headerLine)
...
 
-        // Check for address, and set if found
-        if (preg_match('/^(?P<name>.*?)<(?P<email>[^>]+)>$/', $value, $matches)) {
-            $senderName = trim($matches['name']);
-            if (empty($senderName)) {
-                $senderName = null;
-            }
-            $senderEmail = $matches['email'];
+        if ($hasMatches !== 1) {
+            throw new Exception\InvalidArgumentException('Invalid header value for Sender string');
```
Maks3w: I think we should use a custom exception like MalformedHeaderException instead InvalidArgumentException

For me InvalidArgumentExceptions never should be catchable and means the develop is using wrong the API.

This case does not mean develop is using wrong the class. Developer cannot fully control the input and may is wrong in his/her data source.

test/Header/SenderTest.php
```diff
@@ -190,4 +190,50 @@ public function invalidSenderEncodedDataProvider()
...
+
+        if ($expectedException) {
+            return;
+        }
+
+        $this->assertSame($expectedName, $header->getAddress()->getName());
+        $this->assertSame($expectedEmail, $header->getAddress()->getEmail());
+    }
+
+    public function headerLinesProvider()
```
Maks3w: Split the data provider in two valid data and invalid data

src/Header/Sender.php
Maks3w: Also HeaderInterface docblock should document this type of exception scenarios

test/Header/SenderTest.php
```diff
@@ -190,4 +190,50 @@ public function invalidSenderEncodedDataProvider()
...
+    {
+        $mailInvalidArgumentException = 'Zend\Mail\Exception\InvalidArgumentException';
+        $headerInvalidArgumentException = 'Zend\Mail\Header\Exception\InvalidArgumentException';
+
+        return array(
```
Maks3w: now is mandatory use short array syntax. Next push will be checked against this new guidelines.

test/Header/SenderTest.php
```diff
@@ -190,4 +190,50 @@ public function invalidSenderEncodedDataProvider()
...
+            // [ header line, expected sender name, expected email address, expected exception class ]
+            array( 'Sender: foo@bar', null, 'foo@bar'),
```
Maks3w: If possible add a description to each data case. Descriptions are added as array keys

```
'Testing <> syntax' => ['Sender: <.....>'],
```

src/Header/Sender.php
```diff
@@ -39,23 +39,23 @@ public static function fromString($headerLine)
...
         if (strtolower($name) !== 'sender') {
-            throw new Exception\InvalidArgumentException('Invalid header line for Sender string');
+            throw new Exception\InvalidArgumentException('Invalid header name for Sender string');
         }
 
-        $header      = new static();
-        $senderName  = '';
-        $senderEmail = '';
+        $header     = new static();
+        $hasMatches = preg_match('/^(?:(?P<name>.+)\s)?(?(name)<|<?)(?P<email>[^\s]+?)(?(name)>|>?)$/', $value, $matches);
```
Maks3w: Please add a comment describing how the regex works. Sometimes is very difficult to understand what the regex do.

Maks3w: For example put the relation between each group and the ABNF

src/Header/Sender.php
Maks3w: /cc @weierophinney

test/Header/SenderTest.php
```diff
@@ -190,4 +190,50 @@ public function invalidSenderEncodedDataProvider()
...
+            array( 'Sender: <foo@bar>', null, 'foo@bar'),
+            array( 'Sender:    foo@bar', null, 'foo@bar'),
+            array( 'Sender: name <foo@bar>', 'name', 'foo@bar' ),
+            array( 'Sender: <weird name> <foo@bar>', '<weird name>', 'foo@bar' ),
+            array( 'Sender: moar words <foo@bar>', 'moar words', 'foo@bar' ),
+            array( 'Sender: foo', null, null, $mailInvalidArgumentException ),
```
Maks3w: This test cases should be added to invalidSenderEncodedDataProvider

Maks3w:
@stefanotorresi Do you will complete this?

stefanotorresi:
@Maks3w sorry I couldn't attend to this for a while, but I'll try and get at it by tomorrow, otherwise I'll let you know.

stefanotorresi:
@Maks3w should be good enough now.
I overlooked your proposal for the custom exception because I think it should be implemented in a separate PR, and for every other header as well.

test/Header/SenderTest.php
```diff
@@ -204,4 +204,61 @@ public function invalidSenderEncodedDataProvider()
...
+    public function testFromStringWithInvalidInput($headerString, $expectedException, $expectedMessagePart = '')
+    {
+        $this->setExpectedException($expectedException, $expectedMessagePart);
+
+        Header\Sender::fromString($headerString);
+    }
+
+    public function invalidHeaderLinesProvider()
+    {
+        $mailInvalidArgumentException = 'Zend\Mail\Exception\InvalidArgumentException';
```
Maks3w: You can replace these variables with php 5.5 class constant

Maks3w:
Could you reuse existing data providers?

stefanotorresi:
> Could you reuse existing data providers?

@Maks3w not without making the tests much less clear.

stefanotorresi:
@Maks3w to be clear, this is good to merge AFAIC, do you need me to do something?

stefanotorresi:
@weierophinney is there anything I can do to get this merged? I saw two releases recently but this has slipped under the radar.

PR #10, opened 2015-06-04
remove $self = $this for PHP >= 5.5

samsonasik:


samsonasik:
rebased, should be merge-able now

PR #11, opened 2015-06-09
Fix mail sender

denis-sokolov:
If I'm not mistaken, in RFC 5322 [3.6.2.](https://tools.ietf.org/html/rfc5322#section-3.6.2) states `sender = "Sender:" mailbox CRLF`, and [3.4.](https://tools.ietf.org/html/rfc5322#section-3.4) `mailbox = name-addr / addr-spec`, where `addr-spec = local-part "@" domain`.
More specifically, I got an email sent from Gmail with `Sender: foo@bar`.

### DIFF
--- a/src/Header/Sender.php
+++ b/src/Header/Sender.php
@@ -53,6 +53,8 @@ class Sender implements HeaderInterface
                 $senderName = null;
             }
             $senderEmail = $matches['email'];
+        } else {
+            $senderEmail = $value;
         }
 
         $header->setAddress($senderEmail, $senderName);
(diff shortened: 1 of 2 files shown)


mwillbanks:
@denis-sokolov Thanks for this PR; however, you did modify the test behavior and as such the conditions from the previous test also need to continue in being tested.

denis-sokolov:
Thank you for your note.
I may misunderstand, but perhaps you missed that [I do replay](https://github.com/zendframework/zend-mail/pull/11/files#diff-4ec1c1325f86557a707e909689c2df26R152) existing tests using `array_merge`.

mwillbanks:
:bulb: 
missed that line!

PR #17, opened 2015-06-17
Allow horizontal tab as valid header field body

FraGoTe:


Maks3w:
Need unit tests

weierophinney:
Absolutely not.

This functionality exists to close [ZF2015-04](http://framework.zend.com/security/advisory/ZF2015-04), which exists to prevent message splitting. [RFC 2822](https://www.ietf.org/rfc/rfc2822.txt) clearly outlines what characters are valid for header values, and the horizontal tab is **not** in that list.

As soon as we start selectively allowing characters disallowed by the spec, we begin opening ourselves to potential security vectors.

Maks3w:
This is related with https://github.com/zendframework/zf2/issues/7591

weierophinney:
Based on zendframework/zf2#7591, this is specific to the `Received` header. The general ABNF for headers does *not* allow whitespace except as part of a header folding sequence (which we already detect and handle). (See [section 2.2](http://tools.ietf.org/html/rfc2822#section-2.2) for details).

As such, the appropriate fix is to provide custom validations inside of `Zend\Mail\Header\Received`, and not create a general change that affects all headers.

weierophinney:
@FraGoTe I'm willing to do entertain the change, but it has to be specifically in the Received header implementation; I don't want to introduce new potential vectors in other headers.

FraGoTe:
@weierophinney create a custom validation in Zend\Mail\Header\Received it is ok, I think.

dendemann:
Thanks for discussing this one. RFC2822 was obsoleted by RFC5322 in 2008. [Section 2.2](http://tools.ietf.org/html/rfc5322#section-2.2) was revised so HTABs in header field bodies seem to be allowed in general since then.

PR #16, opened 2015-06-17
Use different encoding for ach different Header

leonardo-piseri:
This fixes issue #15 and maybe removes the need for pull request #1 
Actually I can't understand the meaning of the attribute "encoding" for Headers ( as collection of Header ). 
Each header can have a different encoding. Header::getEncoding() actually sets the encoding for the header.

PR #18, opened 2015-06-19
Handle non ASCII headers

leonardo-piseri:
See closed pull request #16 
The new code should now pass tests.

### DIFF
--- a/src/Headers.php
+++ b/src/Headers.php
@@ -152,7 +152,7 @@ class Headers implements Countable, Iterator
     {
         $this->encoding = $encoding;
         foreach ($this as $header) {
-            $header->getEncoding();
+            ($encoding === 'ASCII') ? $header->getEncoding() : $header->setEncoding($encoding);
         }
         return $this;
     }
--- a/src/Headers.php
+++ b/src/Headers.php
@@ -152,7 +152,7 @@ class Headers implements Countable, Iterator
     {
         $this->encoding = $encoding;
         foreach ($this as $header) {
-            $header->setEncoding($encoding);
+            $header->getEncoding();
         }
         return $this;
     }
--- a/test/HeadersTest.php
+++ b/test/HeadersTest.php
@@ -424,4 +424,22 @@ class HeadersTest extends \PHPUnit_Framework_TestCase
         $headers->addHeaders(['Fake' => ["foo-bar\r\n\r\nevilContent"]]);
         $headers->forceLoading();
     }
+    
+    public function testCanConvertNonASCIIHeaderToStringAfterAddingHeaderToMessage()
+    {
+        //non ASCII characters
+        $str = 'òàù';
+        $header = new \Zend\Mail\Header\GenericHeader();
+        $message = new \Zend\Mail\Message();
+        $headers = new \Zend\Mail\Headers();
+        
+        $header->setFieldName('X-Test');
+        $header->setFieldValue($str);
+        $headers->addHeader($header);
+        $message->setHeaders($headers);
+        $headers->setEncoding('UTF-8');
+        
+		$this->assertTrue(is_string($header->toString()), 'Can call Header::toString()');
+    }
+
 }


src/Headers.php
```diff
@@ -152,7 +152,7 @@ public function setEncoding($encoding)
     {
         $this->encoding = $encoding;
         foreach ($this as $header) {
-            $header->setEncoding($encoding);
+            ($encoding === 'ASCII') ? $header->getEncoding() : $header->setEncoding($encoding);
```
Maks3w: why call `getEncoding`? breaks the lazy load idea

test/HeadersTest.php
```diff
@@ -424,4 +424,22 @@ public function testHeaderCrLfAttackAddHeadersMultiple()
         $headers->addHeaders(['Fake' => ["foo-bar\r\n\r\nevilContent"]]);
         $headers->forceLoading();
     }
+    
+    public function testCanConvertNonASCIIHeaderToStringAfterAddingHeaderToMessage()
+    {
+        //non ASCII characters
+        $str = 'òàù';
+        $header = new \Zend\Mail\Header\GenericHeader();
+        $message = new \Zend\Mail\Message();
```
Maks3w: What is the purpose of this object in the test?

test/HeadersTest.php
```diff
@@ -424,4 +424,22 @@ public function testHeaderCrLfAttackAddHeadersMultiple()
         $headers->addHeaders(['Fake' => ["foo-bar\r\n\r\nevilContent"]]);
         $headers->forceLoading();
     }
+    
+    public function testCanConvertNonASCIIHeaderToStringAfterAddingHeaderToMessage()
+    {
+        //non ASCII characters
+        $str = 'òàù';
+        $header = new \Zend\Mail\Header\GenericHeader();
```
Maks3w: Zend\Mail namespace is already imported. Please use it.

test/HeadersTest.php
```diff
@@ -424,4 +424,22 @@ public function testHeaderCrLfAttackAddHeadersMultiple()
...
+        $message = new \Zend\Mail\Message();
+        $headers = new \Zend\Mail\Headers();
+        
+        $header->setFieldName('X-Test');
+        $header->setFieldValue($str);
+        $headers->addHeader($header);
+        $message->setHeaders($headers);
+        $headers->setEncoding('UTF-8');
+        
+		$this->assertTrue(is_string($header->toString()), 'Can call Header::toString()');
```
Maks3w: 1. assertTrue(is_Foo) can be replaced with assertInternalType(foo)
2. If test is about an encoding problem is better to test output value is exactly the expected encoded value.

src/Headers.php
```diff
@@ -152,7 +152,7 @@ public function setEncoding($encoding)
     {
```
Maks3w: I started to think all this function is a bad solution for proper header encoding. Header encoding should be done by each particular case.

Could we deprecate/remove this function? /cc @weierophinney

test/HeadersTest.php
leonardo-piseri: The problem is not an incorrect encoding, but just the ability to add non ascii headers to a mail message.  The goal of the test is to show that If you create a Zend\Mail\Header() with non ascii value, you can't call toString Method after you add it to a Zend\Mail\Message().
I really don't have good experience in writing tests, I know!

weierophinney:
Per @Maks3w, I'm unsure the test is actually testing the behavior you intend. One way to clarify it is to write it out in plain language first, and then try and write a test to exercise the behavior:

> - When I ...
> - Then I expect ...

Right now, looking at the test, I'm unsure what behavior is being tested; I'd assume that `Header::toString()` would return a string already; what might be more interesting is asserting against the contents of the string — and/or commenting that prior to the fix, an exception occurred at that time. (Assertions against the string contents would still make sense, however!)

Maks3w:
Closed due inactivity

Issue #22, opened 2015-07-29
Subject header is duplicated


matbech:
**Summary**
When sending an mail message with the Sendmail transport, the subject header is duplicated.

**Steps to reproduce**
To reproduce the bug the sample from here can be used:
http://framework.zend.com/manual/current/en/modules/zend.mail.introduction.html

The resulting mail headers look something like this:
Subject: Test Subject
Date: Wed, 29 Jul 2015 17:33:17 +0000
From: =?UTF-8?Q?test?= <test@test.com>
To: test@test.com
Subject: Test Subject

**Cause**
In the Sendmail transport all headers (including the Subject) and the subject itself are passed to the php mail function:
$result = mail($to, $subject, $message, $headers);
The mail function then appends to the $subject to the $headers. I do not think the php mail function should handle this case because it doesn't know what subject the caller intended to use. However I believe php mail should at least issue a warning. 

**Solution**
Pass the headers without the Subject header to the php mail function

arueckauer:
I was not able to reproduce this error. It seems, it is fixed in the current version.

Successfully tested on Windows 10 Pro with php-7.1.26-Win32-VC14-x64, php-7.2.15-Win32-VC15-x64 and php-7.3.2-Win32-VC15-x64.

Used quickstart example for [Sendmail Transport Usage](https://docs.zendframework.com/zend-mail/transport/intro/#sendmail-transport-usage)

```php
use Zend\Mail\Message;
use Zend\Mail\Transport\Sendmail as SendmailTransport;

$message = new Message();
$message->addTo('matthew@example.org');
$message->addFrom('ralph@example.org');
$message->setSubject('Greetings and Salutations!');
$message->setBody("Sorry, I'm going to be late today!");

$transport = new SendmailTransport();
$transport->send($message);
```

Received mail

```
Date: Tue, 05 Mar 2019 08:50:08 +0100
From: ralph@example.org
To: matthew@example.org, matthew@example.org
Subject: Greetings and Salutations!, Greetings and Salutations!

Sorry, I'm going to be late today!
```

glensc:
But why is `To` duplicated in your test result?

And for clarity include the value of your current zend-mail version.

arueckauer:
Very good observation, I will have a look into it. This might be a new issue.

michalbundyra:
This repository has been closed and moved to laminas/laminas-mail; a new issue has been opened at https://github.com/laminas/laminas-mail/issues/72.

PR #23, opened 2015-07-29
Fix issue #22

matbech:
Remove the subject header otherwise it will be duplicated.

### DIFF
--- a/src/Transport/Sendmail.php
+++ b/src/Transport/Sendmail.php
@@ -217,15 +217,14 @@ class Sendmail implements TransportInterface
      */
     protected function prepareHeaders(Mail\Message $message)
     {
-        // On Windows, simply return verbatim
-        if ($this->isWindowsOs()) {
-            return $message->getHeaders()->toString();
-        }
-
-        // On *nix platforms, strip the "to" header
+        // strip the "subject" header
         $headers = clone $message->getHeaders();
-        $headers->removeHeader('To');
         $headers->removeHeader('Subject');
+
+        // On *nix platforms, strip the "to" header as wll
+        if (!$this->isWindowsOs()) {
+            $headers->removeHeader('To');
+        }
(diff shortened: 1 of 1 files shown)


mwillbanks:
@matbech can you please add tests for this?

matbech:
The Subject header is added by the php mail function on Windows. If an extra Subject header is present in the header array the result is a message with two Subject headers.

The request for an automated test is not reasonable because: 
1. There is already enough evidence of the bug:
- Manual test (reason for this PR)
- php source where the "Subject" header is added:
https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/win32/sendmail.c#L595
https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/win32/sendmail.c#L702

2. Writing an STMP server to intercept the mail communication from Window's php:mail is unreasonable, considering the evidence already presented.

glensc:
@matbech can you just make test that if OS is windows the header is removed? not sure if the isWindows result can be mocked.

matbech:
I'm sorry but I have a hard time to understand your sentence. This has been tested on Windows with PHP 7. There is no good reason to leave this bug open for more than 6 months now.

glensc:
Did you test all other supported versions too?, PHP < 7

I'm trying to help, i don't care for one bit for windows support personally!

weierophinney:
We've asked for tests, and blocked this for lack of tests; it's not a question of whether or not we want the fix, but wanting to ensure it *stays fixed*.

You can test this by mocking the result of what `isWindowsOs()` returns (potentially by extending the transport to hardcode the value), and checking the message generated to see if the header was removed.

If/when you are able to do that, please open a new pull request; until then, I'm closing this for now.

PR #24, opened 2015-09-08
Fix mail messages with blank line on top

Maks3w:
At the moment of write this patch GMail (a popular mail provider) prepend mail messages with a line with whitespaces

image

This PR remove that line before parse.

### DIFF
--- a/src/Storage/Message.php
+++ b/src/Storage/Message.php
@@ -42,6 +42,8 @@ class Message extends Part implements Message\MessageInterface
             } else {
                 $params['raw'] = stream_get_contents($params['file']);
             }
+
+            $params['raw'] = ltrim($params['raw']);
         }
 
         if (!empty($params['flags'])) {
--- a/test/Storage/MessageTest.php
+++ b/test/Storage/MessageTest.php
@@ -39,41 +39,19 @@ class MessageTest extends \PHPUnit_Framework_TestCase
         $this->fail('no exception raised while loading unknown file');
     }
 
-    public function testIsMultipart()
+    /**
+     * @dataProvider filesProvider
+     */
+    public function testParseFile($params)
     {
-        $message = new Message(['file' => $this->_file]);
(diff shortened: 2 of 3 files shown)


src/Storage/Message.php
```diff
@@ -42,6 +42,8 @@ public function __construct(array $params)
             } else {
                 $params['raw'] = stream_get_contents($params['file']);
             }
+
+            $params['raw'] = ltrim($params['raw']);
```
Maks3w: Should apply the workaround always? (out of the if block)

Maks3w:
Reading issue https://github.com/zendframework/zend-mail/issues/19 seems this is not enough and another alternative should be provided.

test/Storage/MessageTest.php
```diff
@@ -39,41 +39,19 @@ public function testInvalidFile()
...
-    }
-
-    public function testGetDecodedHeader()
-    {
-        $message = new Message(['file' => $this->_file]);
-
-        $this->assertEquals('Peter Müller <peter-mueller@example.com>', $message->from);
-    }
-
-    public function testGetHeaderAsArray()
```
weierophinney: Having these as separate tests ensures that we can have granular tests for each behavior. Pushing all assertions into a single method makes identifying breakages harder, as the first assertion that fails causes any following assertions to never execute, which means we often end up playing whack-a-mole (fix one assertion, only to find the same test fails on an additional assertion). Please re-instate these discrete methods.

weierophinney:
> Reading issue #19 seems this is not enough and another alternative should be provided.

My reading of the two indicates they're completely separate issues.

src/Storage/Message.php
weierophinney: I think it makes sense.

test/Storage/MessageTest.php
weierophinney: I re-instated these when merging.

PR #27, opened 2015-09-15
Allow to handle reuse time limit

Slamdunk:
Some SMTP server like postfix have a reuse time limit:
http://www.postfix.org/postconf.5.html#smtp_connection_reuse_time_limit

After this time, which is non-infinite by default in postfix, **any** command sent will raise a `4.4.2 *host* Error: timeout exceeded` [Exception](https://github.com/zendframework/zend-mail/blob/release-2.5.2/src/Protocol/AbstractProtocol.php#L327), [`QUIT`](https://github.com/zendframework/zend-mail/blob/release-2.5.2/src/Protocol/Smtp.php#L352) and [`RSET`](https://github.com/zendframework/zend-mail/blob/release-2.5.2/src/Protocol/Smtp.php#L310) too.

The only way to keep using the same smtp object instance, is to force the session end by quitting without QUIT message, and `fclose` the stream resource, to immediately reopen it. This is the use case:
```php
$transport = new Smtp();
$transport->send(new Message());

// A long running script, likely a server-side only tool script
sleep(600);

$transport->send(new Message());
```
Since we do not have an internal transport timer between the two calls, we need to track the time spent saving the connect start time, and checking it in the second send call.

By default, the reuse time limit I set is `-1` which corresponds to infinite and mantain backward compatibility.
`0` can be set to restart the connection on every message.
To tackle the postfix configuration, `270` seconds are enough.

As soon as you accept and merge the PR, I'm going to PR against https://github.com/zendframework/zf2-documentation with explanation on why and how.

### DIFF
--- a/src/Protocol/AbstractProtocol.php
+++ b/src/Protocol/AbstractProtocol.php
@@ -59,6 +59,12 @@ abstract class AbstractProtocol
     protected $socket;
 
     /**
+     * Socket connection start time
+     * @var int
+     */
+    protected $startTime;
+
+    /**
      * Last request sent to server
      * @var string
      */
@@ -131,6 +137,16 @@ abstract class AbstractProtocol
      * Concrete adapters for this class will implement their own unique connect scripts, using the _connect() method to create the socket resource.
      */
     abstract public function connect();
+
+    /**
+     * Retrieve the time the socket has started
+     *
+     * @return string
+     */
+    public function getStartTime()
+    {
+        return $this->startTime;
+    }
 
     /**
      * Retrieve the last client request
@@ -213,7 +229,18 @@ abstract class AbstractProtocol
             throw new Exception\RuntimeException('Could not set stream timeout');
         }
 
+        $this->_startTime();
+
         return $result;
+    }
+
+    /**
+     * Mark the socket connection start time
+     *
+     */
+    protected function _startTime()
+    {
+        $this->startTime = time();
     }
 
     /**
@@ -225,6 +252,8 @@ abstract class AbstractProtocol
         if (is_resource($this->socket)) {
             fclose($this->socket);
         }
+
+        $this->startTime = null;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


Slamdunk:
Travis fails with `Array to string conversion` on Php 5.6.5 + PHPUnit 4.8.7, but I don't know why and on my system is passing either with and without coverage clover.

Any idea?

Slamdunk:
Ping @Maks3w :)

Maks3w:
1. There is no need of modify AbstractProcotol.
2. This is very specific scenario and its better if you implement your own class.

Slamdunk:
1. Yup, no need to change AbstractProtocol, we can create a `Protocol\TimeoutSmtp` and a `Transport\TimeoutSmtp` like I did in my production app if you prefer
2. Postfix is pretty common, long running scripts are a bit less common, but still common, I hope more users can vote it

basz:
I'm wondering why this fix was not excepted. Long running processes that sent email are not a 'specific scenario' IMHO, So if this PR fixes it, why not use it?

I have been using zend-mail for years with this problem and have jumped through several hoops to work around it. (crontab, supervisor restarts, queue systems). Every so often I revisit it hoping the latest version would solve the problem.

basz:
I'm wondering why this PR wasn't accepted. I don't agree with "This is very specific scenario and its better if you implement your own class.". Long running PHP process to sent email is not a very specific scenario IMHO.

I have been looking into this issue for years everytime I wasn't happy with the work arounds so a fix would be very welcomed...

Slamdunk:
@basz https://github.com/zendframework/zend-mail/pull/117 and https://github.com/zendframework/zend-mail/pull/131 are enough to handle nicely this behaviour.

I do this in my app flawlessly; delegators of ServerManager help a lot.

Slamdunk:
@basz you can have a look at https://github.com/Slamdunk/zend-mail-extensions for how to do the things.

basz:
thanks! Would you happen to have a config example? @Slamdunk

Slamdunk:
Soon, need same spare time first :/

basz:
Got it to work, no need to do delegator config as your thing does it. Just needed to inject a service manager. It works great! Thanks again.

PR #31, opened 2015-11-03
Zend\Mail\Transport\SendMail throws wrong exceptions.

mikicaivosevic:
Fix wrong namespace.

Fix https://github.com/zendframework/zf2/issues/7639


src/Transport/Sendmail.php
```diff
@@ -12,7 +12,7 @@
 use Traversable;
 use Zend\Mail;
 use Zend\Mail\Address\AddressInterface;
-use Zend\Mail\Exception;
+use Zend\Mail\Transport\Exception;
```
Maks3w: This use is not required. Please remove

PR #34, opened 2015-11-09
Feature/refactoring bug fix array key defining itself

levelfivehub:
@Maks3w  Very minor, but the array key was assigning a value of itself.


levelfivehub:
@Maks3w Had to close, out of date compared to upstream.  Will re-open when in sync.

PR #35, opened 2015-11-09
removed parts array key defining itself

levelfivehub:
@Maks3w Very minor, but the array key was assigning a value of itself.


Maks3w:
Thx. I knew that. Was written for clarity.

PR #37, opened 2015-11-10
Check that $token[0] exists before trying to use it.

ameir:
With Courier IMAP, you will get an error when checking for $token[0] for some responses, so we ensure that it exists first.

### DIFF
--- a/src/Protocol/Imap.php
+++ b/src/Protocol/Imap.php
@@ -183,6 +183,9 @@ class Imap
         $line = rtrim($line) . ' ';
         while (($pos = strpos($line, ' ')) !== false) {
             $token = substr($line, 0, $pos);
+            if (!strlen($token)) {
+                continue;
+            }
             while ($token[0] == '(') {
                 array_push($stack, $tokens);
                 $tokens = [];


src/Protocol/Imap.php
```diff
@@ -183,36 +183,38 @@ protected function _decodeLine($line)
...
-                array_push($stack, $tokens);
-                $tokens = [];
-                $token = substr($token, 1);
-            }
-            if ($token[0] == '"') {
-                if (preg_match('%^\(*"((.|\\\\|\\")*?)" *%', $line, $matches)) {
-                    $tokens[] = $matches[1];
-                    $line = substr($line, strlen($matches[0]));
-                    continue;
+            if (isset($token[0])) {
```
Maks3w: change this as '! isset(){continue}'

ameir:
All set, let me know if you need any other changes made.

ameir:
@Maks3w @weierophinney do you have any concerns with merging this? I have a few more PRs to write, but want to make sure the path is clear to do so.

Maks3w:
Need tests

ameir:
@Maks3w current tests cover this code path, so I think it's fine test-wise.

Maks3w:
If existing test cover this then this PR its not needed as those tests detect the failure and its fixed.

ameir:
I'm not quite sure how to go about this since we're checking a server response.  I ran tests against a real Courier IMAP server, so if `TESTS_ZEND_MAIL_IMAP_HOST` points to one, then you're set.  I don't think it's worth mimicking a server for this, since this is just a small `isset()` that doesn't change logic for other servers at all.  Let me know what you think.

PR #38, opened 2015-11-10
Support UID FETCH.

ameir:
`FETCH`ing based on IDs isn't reliable when the remote state can change, so this PR offers `UID FETCH` support.


Maks3w:
Need tests

PR #39, opened 2015-11-11
Support setting internaldate on IMAP appends.

ameir:
I'm not too sure why this got overlooked before, but `appendMessage()` previously didn't allow the internaldate to be set, although `Zend\Mail\Protocol\Imap\append()` exposes that functionality.  This PR simply allows the internaldate to be set.


Maks3w:
Need tests

weierophinney:
Also needs a rebase before the next push (ping @ameir ).

test/Storage/ImapTest.php
```diff
@@ -513,6 +513,27 @@ public function testAppend()
         $mail->appendMessage('');
     }
 
+    public function testAppendWithDate()
+    {
+        $mail = new Storage\Imap($this->params);
+        $count = $mail->countMessages();
+        $date = date('d-M-Y H:i:s O', strtotime('12/27/2015'));
+
+        $message = '';
```
CarwynNelson: Is `$message = '';` needed? Could it just be like so:

```language:php
$message = "From: me@example.org\r\n";
$message .= "To: you@example.org\r\n";
// etc etc
```

ameir: It definitely could be.  I did it that way to be consistent with previous tests (https://github.com/ameir/zend-mail/blob/05b0e63c8d5c2c534683473be32319da16df26f2/test/Storage/ImapTest.php#L504).  I'm not sure why it was that way, but I didn't want to break out of whatever standard was in place there, if any.

ameir:
@weierophinney I've rebased, as requested.  Let me know if you need any further changes to be made.

src/Storage/Imap.php
```diff
@@ -465,20 +475,20 @@ public function renameFolder($oldName, $newName)
      *     folder is taken
      * @param null|array $flags set flags for new message, else a default set
      *     is used
+     * @param null|string $date set date for new message in the form of
+     *     "d-M-Y H:i:s O", else current time is used
      * @throws Exception\RuntimeException
      */
-    public function appendMessage($message, $folder = null, $flags = null)
+    public function appendMessage($message, $folder = null, $flags = null, $date = null)
```
weierophinney: This is a potential BC break for anybody extending the class. I think the liklihood of a conflict is low (it would only be an issue if the argument is typehinted in the extension), but we'll need to message it in the release notes at the very least.

src/Storage/Imap.php
```diff
@@ -64,6 +64,16 @@ class Imap extends AbstractStorage implements Folder\FolderInterface, Writable\W
     ];
 
     /**
+     * Return access to protocol class
+     *
+     * @return null|\Zend\Mail\Protocol\Imap
+     */
+    public function getProtocol()
```
weierophinney: What purpose does this method serve?

weierophinney:
Closing; author has not responded to feedback or requests for changes.

If you would like to re-open, please address the feedback and re-submit.

Thanks!

PR #40, opened 2015-11-13
Support IMAP EXAMINE().

ameir:
This PR enables support for IMAP EXAMINE() for read-only folder selection.

### DIFF
--- a/src/Storage/Imap.php
+++ b/src/Storage/Imap.php
@@ -372,13 +372,28 @@ class Imap extends AbstractStorage implements Folder\FolderInterface, Writable\W
      * @throws Exception\RuntimeException
      * @throws Protocol\Exception\RuntimeException
      */
-    public function selectFolder($globalName)
+    public function selectFolder($globalName, $examine = false)
     {
         $this->currentFolder = $globalName;
-        if (!$this->protocol->select($this->currentFolder)) {
+        $method = ($examine ? 'examine' : 'select');
+        if (!$this->protocol->$method($this->currentFolder)) {
             $this->currentFolder = '';
             throw new Exception\RuntimeException('cannot change folder, maybe it does not exist');
         }
+    }
+
+    /**
+     * examine given folder
+     *
+     * folder must be selectable!
+     *
+     * @param  Folder|string $globalName global name of folder or instance for subfolder
+     * @throws Exception\RuntimeException
+     * @throws Protocol\Exception\RuntimeException
+     */
+    public function examineFolder($globalName)
+    {
+        $this->selectFolder($globalName, true);
     }
 
     /**
--- a/src/Storage/Imap.php
+++ b/src/Storage/Imap.php
@@ -372,10 +372,10 @@ class Imap extends AbstractStorage implements Folder\FolderInterface, Writable\W
      * @throws Exception\RuntimeException
      * @throws Protocol\Exception\RuntimeException
      */
-    public function selectFolder($globalName, $examine = false)
+    public function selectFolder($globalName, $readOnly = false)
     {
         $this->currentFolder = $globalName;
-        $method = ($examine ? 'examine' : 'select');
+        $method = ($readOnly ? 'examine' : 'select');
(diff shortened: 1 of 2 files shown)


src/Storage/Imap.php
```diff
@@ -354,16 +354,31 @@ public function getFolders($rootFolder = null)
      * @throws Exception\RuntimeException
      * @throws \Zend\Mail\Protocol\Exception\RuntimeException
      */
-    public function selectFolder($globalName)
+    public function selectFolder($globalName, $examine = false)
```
Maks3w: Revert the changes of this method

src/Storage/Imap.php
```diff
@@ -354,16 +354,31 @@ public function getFolders($rootFolder = null)
...
-        if (!$this->protocol->select($this->currentFolder)) {
+        $method = ($examine ? 'examine' : 'select');
+        if (!$this->protocol->$method($this->currentFolder)) {
             $this->currentFolder = '';
             throw new Exception\RuntimeException('cannot change folder, maybe it does not exist');
         }
     }
 
     /**
+     * examine given folder
```
Maks3w: Add a more detailed description about why this method is different from selectFolder

Maks3w:
Need tests

src/Storage/Imap.php
ameir: Would you rather have `examineFolder()` be identical to `selectFolder()` aside from the examine/select?  i.e. copy/pasting `selectFolder()` and using examine instead?

src/Storage/Imap.php
ameir: There's nothing better than quoting the RFC:

```
The EXAMINE command is identical to SELECT and returns the same
      output; however, the selected mailbox is identified as read-only.
      No changes to the permanent state of the mailbox, including
      per-user state, are permitted; in particular, EXAMINE MUST NOT
      cause messages to lose the \Recent flag.
```

Anyone who uses it should know what it is, just as they know when to use `SELECT`.

weierophinney:
Needs a rebase, @ameir — as well as tests.

ameir:
I added some basic tests to this.  Let me know how it looks.

test/Storage/ImapTest.php
```diff
@@ -594,6 +594,13 @@ public function testSelect()
         $this->assertEquals($status['exists'], 7);
     }
 
+    public function testStorageSelect()
+    {
+        $mail = new Storage\Imap($this->params);
```
Maks3w: You could inject a protocol and spy the expected method is invoked

ameir: Do you have an example of that somewhere for reference?

Maks3w: https://phpunit.de/manual/current/en/test-doubles.html

https://github.com/zendframework/zend-mail/blob/51f5f729c541adc18dfa6497a45be32d1a3d63ae/test/Header/HeaderWrapTest.php#L24-L27

src/Storage/Imap.php
```diff
@@ -372,16 +372,31 @@ public function getFolders($rootFolder = null)
      * @throws Exception\RuntimeException
      * @throws Protocol\Exception\RuntimeException
      */
-    public function selectFolder($globalName)
+    public function selectFolder($globalName, $readOnly = false)
```
weierophinney: Potential BC break for any extensions to this class, but only if those extensions are typehinting the additional, optional argument. Will need to draw attention to this in the release notes.

src/Storage/Imap.php
weierophinney: Please add a `@see <link to RFC section>` to the docblock, and/or quote the RFC to make this clear.

test/Storage/ImapTest.php
weierophinney: Please incorporate the feedback from @Maks3w

weierophinney:
Closing; author has not incorporated requests for changes.

If you would like to re-open, please incorporate the feedback and re-submit.

Thanks!

PR #41, opened 2015-11-13
Support returning IMAP delimiter.

ameir:
This PR exposes the IMAP delimiter character, which is important for many use cases.

### DIFF
--- a/src/Storage/Imap.php
+++ b/src/Storage/Imap.php
@@ -30,6 +30,12 @@ class Imap extends AbstractStorage implements Folder\FolderInterface, Writable\W
     protected $currentFolder = '';
 
     /**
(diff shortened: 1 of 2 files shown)


src/Storage/Imap.php
```diff
@@ -30,6 +30,12 @@ class Imap extends AbstractStorage implements Folder\FolderInterface, Writable\W
     protected $currentFolder = '';
 
     /**
+     * IMAP folder delimiter character
+     * @var string
```
Maks3w: This is null too `string|null`

Maks3w:
Need tests

weierophinney:
Merged to develop for release with 2.7.0.

PR #42, opened 2015-11-27
Inverse decode-split order, allow special char containing labels

mleko:


Maks3w:
Please undo every change is not related with the fix.

test/Header/AddressListHeaderTest.php
```diff
@@ -210,4 +210,27 @@ public function getAddressListsWithGroup()
             ['To: friends: john@example.com; enemies: john@example.net, bart@example.net;', 3, 'john@example.net'],
         ];
     }
+
+    public function specialCharHeaderProvider()
+    {
+        return [
+            ["To: =?UTF-8?B?dGVzdCxsYWJlbA==?= <john@example.com>, john2@example.com", ['john@example.com' => 'test,label', 'john2@example.com'], 'UTF-8'],
```
Maks3w: This is not legible. Split in different lines. Also I suggest assert exactly the same unencoded value.

test/Header/AddressListHeaderTest.php
```diff
@@ -210,4 +210,27 @@ public function getAddressListsWithGroup()
...
+            ['To: "TEST\",QUOTE" <john@example.com>, john2@example.com', ['john@example.com' => 'TEST",QUOTE', 'john2@example.com'], 'ASCII']
+        ];
+    }
+
+    /**
+     * @param string $headerLine
```
Maks3w: Remove this tags. Are noisy.

test/Header/AddressListHeaderTest.php
```diff
@@ -210,4 +210,27 @@ public function getAddressListsWithGroup()
...
+     * @param string $encoding
+     * @dataProvider specialCharHeaderProvider
+     */
+    public function testDeserializationFromSpecialCharString($headerLine, $expected, $encoding)
+    {
+        $header = To::fromString($headerLine);
+        $to = new To();
+        $to->getAddressList()->addMany($expected);
+        $to->setEncoding($encoding);
+        $this->assertEquals($to, $header);
```
Maks3w: Assert the address list contains the expected values.

PR #45, opened 2015-12-02
Use RFC compliant EOL to allow new lines in message body

mleko:
http://www.ietf.org/rfc/rfc2821.txt


test/MessageTest.php
```diff
@@ -688,7 +688,7 @@ public function testRestoreFromSerializedString()
         $this->message->addFrom('matthew@example.com', "Matthew Weier O'Phinney");
         $this->message->addCc('zf-contributors@example.com', 'ZF Contributors List');
         $this->message->setSubject('This is a subject');
-        $this->message->setBody('foo');
+        $this->message->setBody("foo\n\ntest");
```
weierophinney: Changing existing tests can lead to cases where we change existing behavior. As such, I'm going to separate this into its own test when I merge, to verify (a) the old behavior, and (b) the fix this patch presents.

PR #46, opened 2015-12-07
Use consistent EOL, file uses (\r\n)

mleko:


Maks3w:
fixture `mail.txt` has CRLF line terminators so this change is OK.

PR #47, opened 2015-12-07
Removed the zend-servicemanager usage

ezimuel:
This PR removes the usage of zend-servicemanager using a custom SmtpPluginManager. This removal simplify the usage of the component.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,15 +14,14 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-crypt": "~2.5",
+        "zendframework/zend-crypt": "dev-develop as 2.6.0",
         "zendframework/zend-loader": "~2.5",
         "zendframework/zend-mime": "~2.5",
-        "zendframework/zend-stdlib": "~2.5",
+        "zendframework/zend-stdlib": "~2.7",
         "zendframework/zend-validator": "~2.5"
     },
     "require-dev": {
         "zendframework/zend-config": "~2.5",
-        "zendframework/zend-servicemanager": "~2.5",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "~4.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -14,11 +14,12 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-crypt": "dev-develop as 2.6.0",
+        "zendframework/zend-crypt": "^2.5 || ^3.0",
         "zendframework/zend-loader": "~2.5",
         "zendframework/zend-mime": "~2.5",
         "zendframework/zend-stdlib": "~2.7",
-        "zendframework/zend-validator": "~2.5"
+        "zendframework/zend-validator": "~2.5",
+        "container-interop/container-interop": "^1.1"
     },
     "require-dev": {
         "zendframework/zend-config": "~2.5",
--- a/src/Protocol/SmtpPluginManager.php
+++ b/src/Protocol/SmtpPluginManager.php
@@ -9,13 +9,15 @@
 
 namespace Zend\Mail\Protocol;
 
+use Interop\Container\ContainerInterface;
+
 /**
  * Plugin manager implementation for SMTP extensions.
  *
  * Enforces that SMTP extensions retrieved are instances of Smtp. Additionally,
  * it registers a number of default extensions available.
  */
-class SmtpPluginManager
+class SmtpPluginManager implements ContainerInterface
 {
     /**
      * Default set of plugins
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -14,15 +14,14 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-crypt": "~2.5",
+        "zendframework/zend-crypt": "dev-develop as 2.6.0",
```
weierophinney: If we're removing zend-servicemanager as a dep, the change is essentially backwards compatible.

Couldn't this dependency be `^2.5 || ^3.0`? That would make it possible to keep this component in the 2.X series.

src/Protocol/SmtpPluginManager.php
```diff
@@ -9,22 +9,20 @@
 
 namespace Zend\Mail\Protocol;
 
-use Zend\ServiceManager\AbstractPluginManager;
```
weierophinney: Let's have it typehint on container-interop and implement that interface; that'll give the component broader compatibility.

weierophinney:
With the suggested changes, we can target 2.6.0 instead. :smile:

src/Protocol/SmtpPluginManager.php
ezimuel: @weierophinney the get() definition in SmtpPluginManager has the optional parameter $option, that is not defined by container-interop. I know that in PHP the interface compatibility is ok with optional parameters but I want to be sure that we all agree on this "hack".

gianarb:
:+1: :+1:

src/Protocol/SmtpPluginManager.php
weierophinney: @ezimuel We're already implementing it in the abstract plugin manager implementations; essentially, if the second argument is called, we proxy to the `build()` method (which explicitly defines the argument, though still makes it optional). As such: go for it.

ezimuel:
@weierophinney the PR is ready to be merged.

composer.json
```diff
@@ -14,15 +14,15 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-crypt": "~2.5",
+        "zendframework/zend-crypt": "^2.5 || ^3.0",
```
Maks3w: `||` ? Should be a simple `|`

weierophinney: @Maks3w the Composer documentation for that is inconsistent, and uses both. As the parser accepts either, I think using `||` makes sense, for its affiliation with the language itself.

Slamdunk:
@weierophinney AFAIK Zend Framework is following SemVer, isn't it?

Because I just updated my dependancies, and everything broke up because API totally changed in a non BC way.

This PR should be tagged 3.0.0, not 2.6.0, like what it was done to other 2.6 components.

weierophinney:
@Slamdunk What breakage are you seeing in this component, specifically?

Slamdunk:
The fact that all the public method of `AbstractPluginManager` and `ServiceManager` are now missing.

Or in the SemVer the API are only the ones listed in an Interface, in this case `ServiceLocatorInterface`?

weierophinney:
@Slamdunk You're missing the point of my question: *has any code of yours specifically broken?* If so, what specifically? Any code snippets of something *you were actually using that no longer works* would be useful.

The plugin manager implementation in this component was by-and-large an implementation detail, which is why we felt it was safe to refactor it to be standalone; however, I'd like to see if people were relying on the fact that it was an `AbstractPluginManager` implementation specifically, and, if so, how they were using it.

Slamdunk:
Let me preface this comment by saying that refactoring my code to this PR was a mattar of seconds; my speech was more a matter of theory on SemVer than a complain.

As posted on https://github.com/zendframework/zend-mail/pull/27 I need a custom `Transport\Smtp` and `Protocol\Smtp` to face the timeout issues on long running scripts.
So I used to pull the `Protocol\SmtpPluginManager` and call the [setInvokableClass](https://github.com/zendframework/zend-servicemanager/blob/release-3.0.3/src/ServiceManager.php#L395) to use my custom implementation of `Protocol\Smtp`.

After the update this behaviour produced a `Fatal error`, and now I'm forced to subclass the `Protocol\SmtpPluginManager`.

SemVer says:
> 1. MAJOR version when you make incompatible API changes
2. MINOR version when you add functionality in a backwards-compatible manner

Well, a `Fatal error` from `2.5.*` to `2.6.*` it not seems a backwards-compatible change :grin:

weierophinney:
@Slamdunk You've provided me the information I was looking for: you *were* using the features inherited from `AbstractPluginManager`, which means that while we were treating it as an *internal implementation detail*, users were consuming it as part of the public API. As such, yes, this is a BC break, and we need to revisit it.

Thanks for providing me the usage detail!

weierophinney:
@Slamdunk If you update to zend-mail 2.6.1, we've restored the `AbstractPluginManager` implementation.

PR #60, opened 2016-01-22
Fix issue #55

diegosurita:


doc/book/zend.mail.introduction.md
```diff
@@ -26,7 +26,7 @@ $transport->send($mail);
 ```
 
 > ## Note
-#### Minimum definitions
+### Minimum definitions
```
weierophinney: On the off-chance you do more PRs for headers... 

When you see `> ## Note` leading into a header, the header and the paragraph(s) under it all need to also be prefixed by `> `; that way they will be rendered as notes in the documentation.

I'll take care of the ones in this PR; thanks a ton!

PR #61, opened 2016-01-30
add format param to toArray

glensc:
this allows exporting headers in raw or encoded form

### DIFF
--- a/test/HeadersTest.php
+++ b/test/HeadersTest.php
@@ -309,6 +309,31 @@ class HeadersTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals($expected, $string);
     }
 
+    /**
+     * @test that toArray can take format parameter
+     * @link https://github.com/zendframework/zend-mail/pull/61
+     */
+    public function testToArrayFormat()
+    {
+        $raw_subject = '=?ISO-8859-2?Q?PD=3A_My=3A_Go=B3?= =?ISO-8859-2?Q?blahblah?=';
+        $headers = new Mail\Headers();
+        $subject = Header\Subject::fromString("Subject: $raw_subject");
+        $headers->addHeader($subject);
+        // default
+        $array = $headers->toArray(Header\HeaderInterface::FORMAT_RAW);
+        $expected = [
+            'Subject' => 'PD: My: Gołblahblah',
+        ];
+        $this->assertEquals($expected, $array);
+
+        // encoded
+        $array = $headers->toArray(Header\HeaderInterface::FORMAT_ENCODED);
+        $expected = [
+            'Subject' => '=?UTF-8?Q?PD:=20My:=20Go=C5=82blahblah?=',
+        ];
+        $this->assertEquals($expected, $array);
+    }
+
     public static function expectedHeaders()
     {
         return [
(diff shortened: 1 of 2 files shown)


glensc:
is zend-mail no longer maintained? or why such trivial PR is not responded?

froschdesign:
@glensc 
We are currently working on other components.

In the meantime you can add unit tests.

glensc:
and so are other my tickets just quiet:

  - #69 
  - #64 - this is actually blocker why i can't use this library in my project

Maks3w:
Closed due lack of tests

glensc:
test is included

test/HeadersTest.php
```diff
@@ -309,6 +309,31 @@ public function testCastingToStringReturnsAllMultiHeaderValues()
         $this->assertEquals($expected, $string);
     }
 
+    /**
+     * @test that toArray can take format parameter
+     * @link https://github.com/zendframework/zend-mail/pull/61
+     */
+    public function testToArrayFormat()
```
Maks3w: This tests holds two scenarios. One method, one scenario.

Maks3w:
I have concerns about the new argument.

I think is best to declare a new method like toArrayRaw, toArrayEncoded and deprecate old method toArray

Maks3w:
Additionally I have doubts about why this feature is required. Could you elaborate why we should merge this?

glensc:
it's needed when attempting to reconstruct mail object because Message() constructor needs input in raw format not to throw exception.

glensc:
i added as parameter as other methods seemed to have similar interface: `public function getFieldValue($format = HeaderInterface::FORMAT_RAW)`, seemed more consistent.

test/HeadersTest.php
Maks3w: Please split this test

glensc: I was waiting when communication finishes, if the communication ends up whole PR being rejected, i don't see point fixing the tests. so please respond after my replies which changes should be done that you initially mentioned (besides test splitting)

ezimuel:
@Maks3w the PR proposed by @glensc makes sense to me. I'm going to merge it.

ezimuel:
@glensc Thanks for the PR and sorry for the delay!

glensc:
@ezimuel thanks and any pointers how to get this to 2.4 branch?

Maks3w:
This is not a critical bug fix. It won't be backported to 2.4.

glensc:
Yet it the change does not break anything. i need php5.3 support

weierophinney:
@glensc The release process for 2.4 requires a release of the entire framework, which is why we only do so for critical or security fixes.

If you require the functionality in that version, you can also extend the class to add it, and consume your extending class within your own code.

glensc:
@weierophinney extending is not possible usually as hardcoded class names in the framework.

for _example_ i can not override [getFieldValue](https://github.com/zendframework/zend-mail/blob/release-2.5.2/src/Header/GenericHeader.php#L141-L148) because all classes extend from `GenericHeader`, not the class i wrote to extend `GenericHeader`. note this is just example why extending is always not possible to patch functionality.

glensc:
@weierophinney i'd be happy if the change just lands in 2.4 tree, whenever next 2.4 release is actually happening. then the [fork i maintain](https://github.com/glensc/zend-mail/tree/develop-2.4) (for [composer use](https://github.com/eventum/composer/commit/dc42d9ea98be85f53f427f7b7b368dfd20755bc7)) would not diverge from upstream code.

PR #66, opened 2016-02-11
Remove zend-servicemanager dependency

mtymek:
Looks like `zend-servicemanager` is not required by this component - I can't find any references in the code, and tests are not failing without it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
     },
     "require-dev": {
         "zendframework/zend-config": "~2.5",
-        "zendframework/zend-servicemanager": "~2.5",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "~4.0"
     },


vaclavvanik:
Service manager is used in smtp transport

https://github.com/zendframework/zend-mail/blob/master/src/Transport/Smtp.php#L358

mtymek:
Indeed, how could I missed that! Thanks for pointing this out :)

Maks3w:
We could consider apply this https://github.com/zendframework/zend-math/pull/5

weierophinney:
@Maks3w In this particular case, the manager may need extensibility, due to variations in authentication protocols between servers. We cover the most common, but I've assisted somebody before in writing custom auth to plugin to the transport in the past. That makes this situation different from zend-math, where there are literally only the two alternatives.

kynx:
Forgive me if I'm wrong, but didn't #47 make the `SmtpPluginManager` standalone? I can't see anywhere else SM is used...

mtymek:
OK, I just realized that I did my PR against develop branch, while I should be using master - that's why I didn't realize `Zend\ServiceManger` is still used. I created new pull request (#68) with v3 compatibility updates.

ezimuel:
@weierophinney actually we removed the zend-servicemanager dependency with #47. People can extend if they want because SmtpPluginManager is compliant with container-interop. I think we can live with that. /cc @kynx, @mtymek

weierophinney:
@ezimuel @kynx — I see we merged that, but the develop branch still contains a listing in `require-dev` for zend-servicemanager. Trying to figure out why...

weierophinney:
Okay, figured it all out. While we've made the standalone version:

- we never released it, as zend-mail has hard-deps on zend-crypt and zend-validator, and we needed to ensure it would work with forwards-compatible versions of these.
- we forgot to remove the zend-servicemanager dev dependency.

I'll create a quick PR to accomplish those things.

PR #68, opened 2016-02-13
ServiceManager v3 compatibility

mtymek:
1. Updated composer dependencies
2. Updated SmtpPluginManager according to https://github.com/zendframework/maintainers/wiki/ZF3-ServiceManager-component-refactors,-phase-2:
  * removed invokables
  * added aliases
  * added factories
  * added v3-compatible `validate()` method
  * update `validatePlugin` to proxy to `validate`
3. Added `PluginManagerCompatibilityTest`
4. Updated `Zend\Mail\Transport\Smtp::getPluginManager()` to inject empty `ServiceManager`, as it is required by AbstractPluginManager constructor.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     },
     "require-dev": {
         "zendframework/zend-config": "~2.5",
-        "zendframework/zend-servicemanager": "~2.5",
+        "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "~4.0"
     },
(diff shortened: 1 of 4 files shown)


gianarb:
:+1:

weierophinney:
@mtymek The point being made on #66 is that in *develop* we've created a standalone implementation; we just hadn't released it yet, due to the fact that some required dependencies (zend-validator in particular) didn't have forwards-compatible versions yet. As such, we need to (against develop):

- update the composer.json to use stable, forwards-compatible versions
- remove zend-servicemanager from the `require-dev` section
- create a 2.6.0 release.

I'll work on an alternate pull request for that. In the meantime, closing this one, as #47 removes the need for compatibility updates.

PR #69, opened 2016-02-15
make zend-crypt optional

glensc:
zend-crypt is rather optional, used in smtp auth only

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,6 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-crypt": "~2.5",
         "zendframework/zend-loader": "~2.5",
         "zendframework/zend-mime": "~2.5",
         "zendframework/zend-stdlib": "~2.5",
@@ -27,6 +26,7 @@
         "phpunit/PHPUnit": "~4.0"
     },
     "suggest": {
+        "zendframework/zend-crypt": "Crammd5 support in SMTP Auth",
         "zendframework/zend-servicemanager": "Zend\\ServiceManager component"
     },
     "minimum-stability": "dev",
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
     },
     "require-dev": {
         "zendframework/zend-config": "~2.5",
+        "zendframework/zend-crypt": "~2.5",
         "zendframework/zend-servicemanager": "~2.5",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "~4.0"


gianarb:
Hello @glensc adds it as require-dev to fix the test suite..

glensc:
tests should pass now, not sure how to force re-run them on travis

ezimuel:
@glensc I re-executed the test on Travis now it's green, thanks!
Note: to restart a build on travis you need admin permission on the project.

Maks3w:
@ezimuel this is a BC Break. You can't change require dependencies.

weierophinney:
@Maks3w I disagree. SMTP authentication is a *subset* of the functionality, and requirements specific to it should have been considered optional from the outset.

ezimuel:
@Maks3w this is not a BC break! The zend-crypt requirement is only for a specific functionality, including it in [require-dev](https://getcomposer.org/doc/04-schema.md#require-dev) and [suggest](https://getcomposer.org/doc/04-schema.md#suggest) doesn't break default installations.

Maks3w:
If you have a default installation and use the SMTP authentication feature next time you do `composer update` zend-crypt may to be removed and next deploy on production will fail because zend-crypt does not exists.

Removing `require` dependencies has the same effect as remove any class on the repo. We all agree any removal is a BC Break, the same is applicable to this.

ezimuel:
@Maks3w only if you add the option `--no-dev` the require-dev packages are removed from the installation ([please read composer doc](https://getcomposer.org/doc/04-schema.md#require-dev)). So the command `composer update` will not remove it. Again, this is not a BC break!

Maks3w:
@ezimuel You are wrong for the following reasons.

1. Dev dependencies are not expected to be deployed on production (--no-dev)
2. Dev dependencies are only installed when required-dev in the root package (application)

smazurov:
I agree with @Maks3w, its a BC break, that being said I appreciate that it's been merged. Is the fact that its a BC break a problem? Tag it as next version up, and everything's gravy :)

glensc:
to add some more oil to fire, it's even subset of SMTP Auth: `Crammd5 support in SMTP Auth`

ezimuel:
@Maks3w let's put in this way, for the policy of the Zend Framework project, this is not a BC break! BC break means code that needs to be changed. This is only a requirement change, people needs (eventually) to execute a composer command to install the require-dev specific packages here. No code change.

Maks3w:
They have to change their `composer.json` and modify the `require` section.

ezimuel:
@Maks3w with code change I mean PHP code. I think we need to discuss this with all the Zend Framework members, we need a common position on the require-dev usage. We cannot have different opinions on this. /cc @weierophinney

Maks3w:
Indeed.  My position is that anyone should be capable of update from one version to any most recent backward compatibility version without change nothing in the environment or the code (except Composer's lock file if exists)

Sorry but this implies PHP version changes or extensions version should be released as a mayor version.

Why we should do this? Because composer update should always install the latest version (except major versions) so anyone feel safe for jump from 1.0.0 to 1.99.99 blindly. This is critical when release security bugfixes. It's insane to think for install next backward compatibility version I must upgrade the server OS, recompile PHP or any other kind of time required and risky task.

We could argue LTS versions have this constraints in consideration but:

- Security issues may exist only for the most recent non LTS versions. 
- Current LTS version is many years old. 
- Any stable version released is valid for deployment in production. Not only LTS versions. 


All these principles applied to the current PR (guessing will be released as a minor or patch version) result in the following chain of events:

- Add zend-mail in the required section of composer.json
- Write unit tests and mock zend-mail following the best practices. 
- Send the  app for deployment. 
- CI green so deploy. 
- Application works fine.
- This PR is released. 
- Now guess the application does not have composer.lock or zend-mail version is updated for any other reason (foo-baz require an updated zend-mail,  composer update,...)
- CI green so deploy. 
- Application blowup. PHP Error Class Zend\Crypt\... not found.

ezimuel:
@Maks3w I can agree but I think you should start to use a language less aggressive. If you use terms like "it's insane" or "you are wrong" in your replies, I don't think this will help the community or the project. The usage of the right words is important and we should promote a more relaxed and civil code review, I think.

Maks3w:
@ezimuel Thx. I'll try to improve my language

weierophinney:
@Maks3w — I both agree and disagree.

When we bump to a new minor version, we are also signaling *new features*. Those may also include dependency changes; in point of fact, we've changed required PHP versions between minor versions several times, and removing or adding dependencies has also been done.

In most cases, one of a few situations will happen when that occurs:

- Composer will refuse to update to the new minor version due to discrepancies in dependencies. This is reasonable; the user then has to opt-in to the new minor version by updating their own dependencies.
- Composer will update to the new version, but code will not work as it had previously due to missing dependencies. This is sub-optimal, but provides us a chance to educate users as to the change and the purpose behind it. This information is usually communicated via the CHANGELOG; if a user raises an issue about it, we can point them to the release notes.
- Users will indicate situations we were unable to foresee, and we will determine we need to rollback the change. This is a situation of last-resort, and has occurred *very* infrequently, though it *has* occurred.

In the case that a user has a `composer.lock` already, `composer install` will install the original dependencies and original versions. The only time the scenario you list above will cause an issue is on a `composer update`; best practices dictate that you should *always* run you project through CI whenever you do an update to detect such issues early. Deploying to production and using `composer update` (or `composer install` without a `composer.lock` file) is a recipe for problems, which is why the `composer.lock` file exists, and why `install` should always be used when readying an application for production.

As such, I definitely refute your assertion that we should not make PHP version or dependency changes in packages unless a new major version is being released. Doing so will stagnate the project and/or lead to version inflation. As a general rule, most such changes are fine for release with minor versions, *unless a user reports a critical issue to the contrary*. **This is, and continues to be our policy,** regardless of how you feel about it personally.

Now, all that said, this particular change, however, is one which is slightly more problematic, as, as you have noted, the zend-crypt functionality is not in a part you can usually mock with tests, which means it's an issue that likely will not be detected until production. As such, I agree that *this particular change* should be rolled back, for the stated reasons.

PR #71, opened 2016-02-19
allow PHPUnit 5.2

remicollet:
I think this will be nice to use upstream maintained version, especially with PHP 7

With this, PHP < 5.6 will use the same version, PHP 5.6 and 7.0 wille use 5.2

Minimal version set to 5.2.9 because of https://github.com/sebastianbergmann/phpunit/pull/2083

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_install:
   - if [[ $EXECUTE_TEST_COVERALLS == 'true' ]]; then composer require --dev --no-update satooshi/php-coveralls ; fi
 
 install:
-  - travis_retry composer install --no-interaction --ignore-platform-reqs
+  - travis_retry composer install --no-interaction
 
 script:
   - if [[ $EXECUTE_TEST_COVERALLS == 'true' ]]; then ./vendor/bin/phpunit --coverage-clover clover.xml ; fi
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require-dev": {
         "zendframework/zend-config": "^2.6",
         "fabpot/php-cs-fixer": "1.7.*",
-        "phpunit/PHPUnit": "~4.0"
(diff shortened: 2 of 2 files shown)


Maks3w:
What is the real benefit? It works with ^4. why maintain test code for two major versions?

:-1:

remicollet:
Because "PHPUnit 4.8 will receive bug fixes until August 7, 2016." See: https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-4.8.0

> why maintain test code for two major versions?

There is absolutely no difference in the tests.

Maks3w:
So they will support v4 until PHP 5.5 reach his EOL.

I don't think we need to upgrade now and we should do it when we really need to do it. This is, if nobody spot a bug in PHPUnit v4 codebase, PHP 7.x incompatibility or lack of feature then we don't need to upgrade.

remicollet:
Please note, this is NOT about update (raising minimal version), but about allowing both version.

And FYI, I run all ZF framework test suite using PHPUnit 5.2, everything is OK.

Maks3w:
If PHPUnit will be a *production* dependency then I agree with to maintain both versions but this is a development dependency.

Nobody want complexities in the development tools as there are too many complexity in the production code. This is just a source of nightmares

sebastianbergmann:
@remicollet From https://phpunit.de/: "PHPUnit 4.8 is the current old stable release series. It became stable on August 7, 2015. Its support ends on February 3, 2017."

@Maks3w PHPUnit 4 is not supported on PHP 7.

Maks3w:
@sebastianbergmann I know but is working https://travis-ci.org/zendframework/zend-mail/jobs/117650944#L148-L333

as I said previously 

> I don't think we need to upgrade now and we should do it when we really need to do it. This is, if nobody spot a bug in PHPUnit v4 codebase, PHP 7.x incompatibility or lack of feature then we don't need to upgrade.

PR #79, opened 2016-03-31
Fix problem with uniqueIds and receiving Messages from Maildir

WemmZi:
When using Zend\Mail\Storage\Writable\Maildir and saving mails to it an receiving the uniqueId, that uniqueId did never match the determined $uniq when reading the mail from Maildir.
uniq should be like ##########.####.####.@@@@@
uniq was like ##########.####.####.@@@@@,S=####

### DIFF
--- a/src/Storage/Maildir.php
+++ b/src/Storage/Maildir.php
@@ -294,7 +294,7 @@ class Maildir extends AbstractStorage
 
             ErrorHandler::start(E_NOTICE);
             list($uniq, $info) = explode(':', $entry, 2);
-            list(, $size) = explode(',', $uniq, 2);
+            list($uniq, $size) = explode(',', $uniq, 2);
             ErrorHandler::stop();
             if ($size && $size[0] == 'S' && $size[1] == '=') {
                 $size = substr($size, 2);


Maks3w:
Need tests

WemmZi:
What kind of tests do you need?

Maks3w:
Every contribution must provide unit tests for check the validity of the changes and avoid regressions.

WemmZi:
I fixed the value of one variable in a protected method. The existing unit tests should be sufficient.

weierophinney:
@WemmZi — 

> I fixed the value of one variable in a protected method. The existing unit tests should be sufficient.

That's a false assumption. The change you made does a new assignment for `$uniq`, which could potentially also lead to a regression (users may have relied on that previously). As such, we need a test demonstrating:

- the code to reproduce the issue
- what the expected behavior is

and that will need to pass *along with the existing unit tests* so we can validate there are no conflicts. (We've observed such situations many times in the past!)

Thanks in advance!

WemmZi:
@weierophinney
Thanks for your wide explanation.

I will see, if i find the time for writing unit tests for that.

If anyone else has the time for writing this unit tests. You are invited to support me. :-)

Before:
1. uniqueId when storing mail in Maildir: ##########.####.####.@@@@@
2. unqiueId when getting UniqueId by Index from Maildir: ##########.####.####.@@@@@,S=####
-> getNumberByUniqueId is not usable, because you never get a match between 1 and 2

After:
1. uniqueId when storing mail in Maildir: ##########.####.####.@@@@@
2. unqiueId when getting UniqueId by Index from Maildir: ##########.####.####.@@@@@
-> getNumberByUniqueId is usable, because known uniqueId is equal to uniqueId when reading in Maildir

weierophinney:
Closing for now, due to inactivity. If somebody would like to pick it up, fork from the branch created by @WemmZi and provide unit tests.

PR #82, opened 2016-04-19
Add return type hints

mleko:


Maks3w:
Please, add the descriptions too as appear on Address https://github.com/zendframework/zend-mail/blob/7a11005711632bed0c568c11f50ee8f84a14485f/src/Address.php#L59-L83

CarwynNelson:
Surely the method names should be enough documentation? This just seems to be re-typing what the method name is already telling us it does.

PR #86, opened 2016-04-28
Load headers eagerly to avoid re-encoding problems with multivalve headers

mleko:


weierophinney:
Thanks for the improvement, @mleko !

PR #65, opened 2016-02-10
Fix HeaderValue throwing an exception on legal characters

avasa:
Original pull request [#24](https://github.com/zendframework/zend-mail/pull/28) created Oct 7 2015.

Author of pull request has not responded since Nov 4 2015.

Can we please get this merged and resolved.

Thanks.

PR #87, opened 2016-05-05
Complete #28

weierophinney:
This patch incorporates the patches from #28, with the following additions:

- Incorporates feedback regarding ASCII range.
- Removes an unnecessary test.
- Adds a test case in the `validateValues()` provider to ensure that `\r\n\t` is recognized as a valid header wrap.
- Updates the code so that this last test case actually passes.

In addition, I was unable to run php-cs-fixer when testing after running `composer update --prefer-lowest`. As such, this patch switches to phpcs, and corrects all CS issues flagged.

Fix #28 
Close #65


Maks3w:
@weierophinney Note this should be backported to 2.4 series

weierophinney:
@Maks3w why? I don't see any discussion around that on the original patch...

Maks3w:
Because this bug was introduced by the zend mail headers security patch which it was applied to 2.4 too

glensc:
hard to see what the changes here are, because phpcs diff in same MR. i attempted to cherry-pick this to 2.4 branch, but seems the changes are already there

40fbcbe08998605ca219ecbc7ac3b0c425aa768c
e1c049e759758f2a1a6875208a7f14923ee5292d

PR #116, opened 2016-10-11
Protocol\AbstractProtocol: verbose errors

Slamdunk:
Recently an update of openssl stopped our app (aright) becase one of our partner had messy servers:

```
$ php -r 'var_dump(stream_socket_client("ssl://hide.me:465", $errorNum, $errorStr));var_dump($errorNum, $errorStr);'

Warning: stream_socket_client(): SSL operation failed with code 1. OpenSSL Error messages:
error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure in Command line code on line 1

Warning: stream_socket_client(): Failed to enable crypto in Command line code on line 1

Warning: stream_socket_client(): unable to connect to ssl://hide.me:465 (Unknown error) in Command line code on line 1
bool(false)
int(0)
string(0) ""
```

But it took us a while to get the right error, because an anonym `Could not open socket` was the only message thrown by the smtp protocol.
Also, many times `$errorNum` and `$errorStr` are empty.

With this PR the error-suppressing operator `@` is deleted and a specific error handler to exception function handles the possible warnings.

### DIFF
--- a/src/Protocol/AbstractProtocol.php
+++ b/src/Protocol/AbstractProtocol.php
@@ -206,7 +206,14 @@ abstract class AbstractProtocol
         $errorStr = '';
 
         // open connection
-        $this->socket = @stream_socket_client($remote, $errorNum, $errorStr, self::TIMEOUT_CONNECTION);
+        set_error_handler(
+            function ($error, $message = '') {
+                throw new Exception\RuntimeException(sprintf('Could not open socket: %s', $message), $error);
+            },
+            E_WARNING
+        );
+        $this->socket = stream_socket_client($remote, $errorNum, $errorStr, self::TIMEOUT_CONNECTION);
+        restore_error_handler();
 
         if ($this->socket === false) {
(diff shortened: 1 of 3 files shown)


weierophinney:
Thanks, @Slamdunk

src/Protocol/AbstractProtocol.php
```diff
@@ -206,7 +206,14 @@ protected function _connect($remote)
         $errorStr = '';
 
         // open connection
-        $this->socket = @stream_socket_client($remote, $errorNum, $errorStr, self::TIMEOUT_CONNECTION);
+        set_error_handler(
+            function ($error, $message = '') {
+                throw new Exception\RuntimeException(sprintf('Could not open socket: %s', $message), $error);
```
glensc: isn't such code faulty that error handler is not restored when exception is thrown?

point being that throwing exception not necessarily terminate the program, it may resume working with this error handler being still set.

PR #123, opened 2016-11-23
Fix documentation error for emails with attachements

MatthiasKuehneEllerhold:
* Add missing " and ; to code example
* add note about the correct order of content mime parts


kynx:
That missing semi-colon just wasted an hour of my life.
👍

PR #124, opened 2016-11-29
Require $this->headers to be instance of Headers

paragonie-scott:
Turns a fatal error into a catchable Exception (which allows users to have a helpful stack trace).

This sort of change will be unnecessary in PHP 7 (hooray return type declarations and catchable `TypeError`s).

### DIFF
--- a/src/Storage/Part.php
+++ b/src/Storage/Part.php
@@ -294,10 +294,16 @@ class Part implements RecursiveIterator, Part\PartInterface
      */
     public function getHeader($name, $format = null)
     {
-        $header = $this->getHeaders()->get($name);
+        $headers = $this->getHeaders();
+        if (!($headers instanceof Headers)) {
+            throw new Exception\RuntimeException(
+                '$this->headers must be an instance of Headers'
+            );
+        }
+        $header = $headers->get($name);
         if ($header === false) {
             $lowerName = strtolower(preg_replace('%([a-z])([A-Z])%', '\1-\2', $name));
-            $header = $this->getHeaders()->get($lowerName);
+            $header = $headers->get($lowerName);
             if ($header === false) {
                 throw new Exception\InvalidArgumentException(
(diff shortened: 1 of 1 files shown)


paragonie-scott:
There may be other areas where it would be beneficial to convert fatal errors to catchable exceptions besides the instance I found here.

Ocramius:
> There may be other areas where it would be beneficial to convert fatal errors to catchable exceptions besides the instance I found here.

Wouldn't bother too much about it, given the upcoming [drop of PHP 5.x support](https://php.net/supported-versions.php)

src/Storage/Part.php
```diff
@@ -294,10 +294,16 @@ public function getHeaders()
      */
     public function getHeader($name, $format = null)
     {
-        $header = $this->getHeaders()->get($name);
+        $headers = $this->getHeaders();
```
Ocramius: The API of [`getHeaders`](https://github.com/paragonie-scott/zend-mail/blob/6a53532344db6159cd894082550a40c2a61fb8d1/src/Storage/Part.php#L268) specifies that a `Headers` instance *MUST* be returned. If that is not true, then `getHeaders` needs fixing, and you are fixing a symptom here.

src/Storage/Part.php
```diff
@@ -383,7 +389,13 @@ public function __get($name)
      */
     public function __isset($name)
     {
-        return $this->getHeaders()->has($name);
+        $headers = $this->getHeaders();
```
Ocramius: The API of [`getHeaders`](https://github.com/paragonie-scott/zend-mail/blob/6a53532344db6159cd894082550a40c2a61fb8d1/src/Storage/Part.php#L268) specifies that a `Headers` instance *MUST* be returned. If that is not true, then `getHeaders` needs fixing, and you are fixing a symptom here.

paragonie-scott:
Right, I'll send a superseding PR.

PR #132, opened 2017-03-02
phpunit 6

kokspflanze:
drop php 5.5
added phpunit 6 support
travis update

i will do the `::class` job after that PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,9 @@ cache:
 
 env:
   global:
-    - SERVICE_MANAGER_VERSION="^3.0.3"
+    - COMPOSER_ARGS="--no-interaction"
+    - COVERAGE_DEPS="satooshi/php-coveralls"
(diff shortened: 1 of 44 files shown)


Xerkus:
@kokspflanze please change branch to develop

kokspflanze:
@Xerkus done

kokspflanze:
@Xerkus ohh didnt saw them, these are skipped atm, so build say nothing=(

changed it=)

about the `Transport\NullTest` these is skipped because it doesnot work on php 7 (Cannot test Null transport under PHP 7; reserved keyword), so dont need to change it, also that namespace doesnt exist on phpunit 5.7.15, so if we change it we need for that the classname hack. i think that is not realy better as the the current solution.

Xerkus:
grep, or its more modern alternative silversearcher, is your friend

Xerkus:
@kokspflanze thank you

PR #128, opened 2017-01-29
added ext-iconv to require

samsonasik:
make it require in the first place to avoid user get error when using it:

```bash
Call to undefined function Zend\Mail\Header\iconv_mime_decode()
```

PR #135, opened 2017-04-04
Update composer.json

glensc:
add missing ext deps for testing

```
Fatal error:  Call to undefined function Zend\Mail\Header\iconv_mime_decode() in src/Header/HeaderWrap.php on line 106
```


Xerkus:
Duplicate of #128

PR #140, opened 2017-05-05
Always escape shell arguments before mail()

paragonie-scott:


Ocramius:
Pretty much agree with the patch, but I'm cracking my whip about test changes that need to be made.

paragonie-scott:
I'm doing it one-file-at-a-time because I'm using github.com, but sure :)

Ocramius:
Some existing tests will also need fixing. I will try prioritizing this today, as it kinda looks bad :|

paragonie-scott:
Yeah :\

Ocramius:
Sorry, didn't get to it due to IRL issues. I'll work on it asap.

Ocramius:
So, I picked this up, and an exception is fired very early on (while setting the `From:` on the `Message` object), so a different string needs to be forged in order to get to a reproducible failure.

As discussed with @paragonie-scott, this is not a security issue report, but just hardening the test suite and improving current reliability against the already reported CVE.

paragonie-scott:
> an exception is fired very early on

That means ZF2 is safe, given a typical usage. This is just defense in depth. :)

Ocramius:
@paragonie-scott the fact that the shell arg escaping is not applied is still bothersome, so the test should indeed verify the exception as it does before paragonie-scott/zend@7551343, but with a different escape sequence (valid domain, valid mail, shell escape).

One possible approach I can think of is to mock the `Message` object (so we verify that even if you want to hurt yourself, the `Sendmail` adapter will not allow you to do so)

paragonie-scott:
I'm really not sure how to do fancy stuff with object mocking in unit tests. :\

It looks like the underlying issue was reported previously as ZF2016-04, but the fix was incomplete.

weierophinney:
Hey, @paragonie-scott — I've pushed a commit to your branch that performs the testing that @Ocramius suggested. If travis looks okay (minus expected issues with the zend-servicemanager compat tests), I'll merge this today.

Thanks!

PR #141, opened 2017-05-05
Ensure getHeaders() always returns an instance of Headers

paragonie-scott:
Supersedes #124


weierophinney:
Thanks, @paragonie-scott

PR #142, opened 2017-05-14
skip test which requires intl extension

glensc:
skip test which requires intl extension, also add such dependency to composer suggest.


glensc:
tests pass locally, those errors happening in travis i've already reported in several other issues/pull-requests.

weierophinney:
Thanks, @glensc

PR #143, opened 2017-05-14
Fix and test ">" being part of email "comment". fixes #127

### LINKED ISSUE #127: Problems with mailadress parsing
Example Email-header:
"Foo <bar" foo.bar@test.com

Description:
The example email-header should be valid according to https://tools.ietf.org/html/rfc2822#section-3.4 but the function AdressList.php/addFromString matches it incorrect. The result has the following form:
"bar <foo.bar@test.com"
This is clearly not a valid adress and therefore causes exceptions in the following code

Solution:
Changing the regex in the addFromString function from lazzy to greedy by removing "?" from the named capture group "name"

- old: ^((?P.*?)<(?P[^>]+)>|(?P.+))$
- new: ^((?P.*)<(?P[^>]+)>|(?P.+))$

glensc:
Fix from https://github.com/zendframework/zend-mail/issues/127

Example Email-header:
```
  "Foo <bar" <foo.bar@test.com>

```
Description:
```
   The example email-header should be valid
   according to https://tools.ietf.org/html/rfc2822#section-3.4
   but the function AdressList.php/addFromString matches it incorrect.
   The result has the following form:
    "bar <foo.bar@test.com"
   This is clearly not a valid adress and therefore causes
   exceptions in the following code

```

### DIFF
--- a/src/AddressList.php
+++ b/src/AddressList.php
@@ -90,7 +90,7 @@ class AddressList implements Countable, Iterator
      */
     public function addFromString($address)
     {
-        if (! preg_match('/^((?P<name>.*?)<(?P<namedEmail>[^>]+)>|(?P<email>.+))$/', $address, $matches)) {
+        if (! preg_match('/^((?P<name>.*)<(?P<namedEmail>[^>]+)>|(?P<email>.+))$/', $address, $matches)) {
             throw new Exception\InvalidArgumentException('Invalid address format');
         }
 
--- a/test/AddressListTest.php
+++ b/test/AddressListTest.php
@@ -14,10 +14,13 @@ use Zend\Mail\AddressList;
 
 /**
  * @group      Zend_Mail
- * @covers Zend\Mail\AddressList<extended>
+ * @covers \Zend\Mail\AddressList<extended>
  */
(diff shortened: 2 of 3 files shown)


glensc:
while writing test, i discovered:

```
        $encodedValue = $to->getFieldValue(Header\HeaderInterface::FORMAT_RAW);
        // FIXME: shouldn't the "name" part be in quotes?
        $this->assertEquals('õlu <bar <foo.bar@test.com>', $encodedValue);
```

this is not good. what place it needs to be fixed?

or it's irrelevant because `FORMAT_RAW` is for display purposes, and irrelevant how it looks?

glensc:
@weierophinney https://github.com/zendframework/zend-mail/pull/143#issuecomment-301312549

can you give some feedback?

Ocramius:
> or it's irrelevant because `FORMAT_RAW` is for display purposes, and irrelevant how it looks?

If this is still relevant, let's push it to a different issue :8ball:

Ocramius:
Merged in zendframework/zend-mail@e75010826bfee8877571b349692cc67361d29351, thanks!

PR #144, opened 2017-05-15
TLS: allow any protocol version

Slamdunk:
Since PHP 5.6.7 the constant `STREAM_CRYPTO_METHOD_TLS_CLIENT` doesn't refer to *any* TLS version, but only to TLS 1.0.

As of yet every zend-mail users are forced to use TLS 1.0 version, and TLS 1.1 and TLS 1.2 are excluded.
We need to readd them manually.

1. https://3v4l.org/llTT8
1. [stream_socket_enable_crypto](https://secure.php.net/manual/en/function.stream-socket-enable-crypto.php) manual page
1. [PHP 5.6.7 revert commit](https://bugs.php.net/bug.php?id=69195)
1. [PHPMailer evidence](https://github.com/PHPMailer/PHPMailer/blob/v5.2.23/class.smtp.php#L354-L370)
1. [New RFC for PHP 7.2](https://wiki.php.net/rfc/improved-tls-constants)

I consider this a **very urgent** fix.

### DIFF
--- a/src/Protocol/Imap.php
+++ b/src/Protocol/Imap.php
@@ -13,6 +13,8 @@ use Zend\Stdlib\ErrorHandler;
 
 class Imap
 {
+    use ProtocolTrait;
+
     /**
      * Default timeout in seconds for initiating session
      */
@@ -102,7 +104,7 @@ class Imap
 
         if ($isTls) {
             $result = $this->requestAndResponse('STARTTLS');
-            $result = $result && stream_socket_enable_crypto($this->socket, true, STREAM_CRYPTO_METHOD_TLS_CLIENT);
+            $result = $result && stream_socket_enable_crypto($this->socket, true, $this->getCryptoMethod());
             if (! $result) {
                 throw new Exception\RuntimeException('cannot enable TLS');
             }
--- a/src/Protocol/Pop3.php
+++ b/src/Protocol/Pop3.php
@@ -13,6 +13,8 @@ use Zend\Stdlib\ErrorHandler;
 
 class Pop3
 {
+    use ProtocolTrait;
+
     /**
      * Default timeout in seconds for initiating session
      */
@@ -113,7 +115,7 @@ class Pop3
 
         if ($isTls) {
             $this->request('STLS');
-            $result = stream_socket_enable_crypto($this->socket, true, STREAM_CRYPTO_METHOD_TLS_CLIENT);
+            $result = stream_socket_enable_crypto($this->socket, true, $this->getCryptoMethod());
             if (! $result) {
                 throw new Exception\RuntimeException('cannot enable TLS');
             }
--- a/src/Protocol/ProtocolTrait.php
+++ b/src/Protocol/ProtocolTrait.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Mail\Protocol;
+
+/**
+ * https://bugs.php.net/bug.php?id=69195
+ */
+trait ProtocolTrait
+{
+    public function getCryptoMethod()
+    {
+        // Allow the best TLS version(s) we can
+        $cryptoMethod = STREAM_CRYPTO_METHOD_TLS_CLIENT;
+
+        // PHP 5.6.7 dropped inclusion of TLS 1.1 and 1.2 in STREAM_CRYPTO_METHOD_TLS_CLIENT
+        // so add them back in manually if we can
+        if (defined('STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT')) {
+            $cryptoMethod |= STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT;
+            $cryptoMethod |= STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT;
+        }
+
+        return $cryptoMethod;
+    }
+}
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -17,6 +17,8 @@ namespace Zend\Mail\Protocol;
  */
 class Smtp extends AbstractProtocol
 {
+    use ProtocolTrait;
+
     /**
      * The transport method for the socket
      *
@@ -176,7 +178,7 @@ class Smtp extends AbstractProtocol
         if ($this->secure == 'tls') {
             $this->_send('STARTTLS');
             $this->_expect(220, 180);
-            if (! stream_socket_enable_crypto($this->socket, true, STREAM_CRYPTO_METHOD_TLS_CLIENT)) {
+            if (! stream_socket_enable_crypto($this->socket, true, $this->getCryptoMethod())) {
                 throw new Exception\RuntimeException('Unable to connect via TLS');
             }
             $this->ehlo($host);
--- a/test/Protocol/ProtocolTraitTest.php
+++ b/test/Protocol/ProtocolTraitTest.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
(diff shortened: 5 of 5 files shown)


Slamdunk:
@weierophinney ping

ezimuel:
@Slamdunk instead of copy the `$cryptoMethod` code in all the classes, we can refactor it using a Trait. Moreover, we need a unit test for that.

Slamdunk:
@ezimuel Ok for centralizing the code, but how to test it?

1. The code contains no logic
2. The test would equal the code itself, I'm not sure something like this would be useful:

```php
class StreamSocketTest exteneds PHPUnit_Framework_TestCase
{
    public function testTlsVersion()
    {
        $cryptoMethod = STREAM_CRYPTO_METHOD_TLS_CLIENT;
        if (defined('STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT')) {
            $cryptoMethod |= STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT;
            $cryptoMethod |= STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT;
        }
        
        $this->assertSame($cryptoMethod, StreamSocket::getCryptoMethod());
    }
}
```

ezimuel:
@Slamdunk for the unit test we should check that using PHP 5.6.7+ the crypto constants are ok. We can use the PHPUnit annotation `@requires PHP 5.6.7` for this test.

Moreover, I suggested to use a [Trait](http://php.net/manual/en/language.oop5.traits.php) and not a static function of a new class. You can use a name like `ProtocolTrait`. Can you change this? Thanks!

Slamdunk:
@ezimuel A test was added in the way that I found the most worthwhile.

I left the code in a new class because traits can easily get bypassed: https://3v4l.org/3DrmE .
The new class express directly the intent, and its behaviour cannot be modified: if the choise of the crypto method must be the same for all the protocols, no protocol should have the chance to alter it.

Do you have particular reasons for using them in this case?

ezimuel:
@Slamdunk the unit test looks perfect! Regarding the Trait I think this is a perfect fit because we just need to share the same code across different classes that does not extends or implements a shared interface. Your point regarding the bypass of the method that's perfectly allowed, it's an issue. If in the future, for any reason, a protocol needs specific configurations it can override it. Moreover, we are providing a Trait for an internal components of a library, that means you should always change the code of the library to override it.

Slamdunk:
@ezimuel I believe the code should protect itself in every scenario: you should be able to change the internal code of a library only if you become aware of every reason behide the already written lines.

With a Trait you are able to modify/override the code and get the tests pass without alteration: this would be a security hole.

Instead, if in the future some evidence will require this code to be altered, you can't skip reading and understanding the current reasons.

ezimuel:
@Slamdunk IMHO I don't see any security issue using a Trait. Again, if someone wants to override the `getCryptoMethod()` he/she needs to change always the source code of the library.
If this is your argument against Trait, when you will use it? I guess never :)

Anyway, I would like to have a different feedback on this, @weierophinney what do you think?

Slamdunk:
> If this is your argument against Trait, when you will use it? I guess never :)

Indeed 👍

weierophinney:
@Slamdunk I'm failing to see how this would present a security issue. Composing a trait is equivalent to compile-time copy-and-paste, and any changes to the trait would be reflected in changes in behavior, and thus _captured in the test suite already_. Using a final static class places a hard dependency within the classes using this instead, and a change to that is _also_ reflected in the test suite results.

My point is: It's not a security issue; it's an issue of preference. As a project, we've generally preferred usage of traits vs static methods for code since we updated to PHP 5.6 as our minimum supported version.

Slamdunk:
@weierophinney The issue here is that we are **not** testing this bug+bugfix.

In order to test this bug we should create a (mock) socket listener and try to connect to it in different cases with different parameters for every `Zend\Mail\Protocol` available.
**IF** we could do this, I would agree with you.

Since we can't do this, suppose the Trait is written and used; it's passed a **year**, and someone creates a PR like this:
```php
// class Smtp
+private function getCryptoMethod()
+{
+    return STREAM_CRYPTO_METHOD_TLS_CLIENT;
+}
+
+public function createTlsSocket($resource)
+{
+    return stream_socket_enable_crypto($resource, true, $this->getCryptoMethod());
+}
```
This would actually break `Smtp::helo` method without notice; all tests will go green because there are no test as of yet that verify that `Smtp::helo` is handling TLSv1.2.
How are you going to spot the error if you don't remember each and every line of the project you are maintaining?

I see only two choises:

1. We create real tests for each `stream_socket_enable_crypto` usage and use-case
1. We enforce with code to think what we are changing

weierophinney:
@Slamdunk Again, in that case, we'd see the addition of the new method during review, giving us time to consider its impact. It's not much different than changing the argument to `stream_socket_enable_crypto()`, which could also happen via a PR, and which would also spawn a serious review. _If_ we have tests in place to verify what arguments are provided to that method, such changes would require accompanying test changes, which would spawn red flags during the review process. Without tests, they will not, even if you have a separate class and method for generating the values.

Which brings me to what I feel is your most important point — that tests do not exist yet to verify that these calls are happening appropriately. This is the crux of the matter, and is not solved with just creating a static class method that you've tested; we need to test that the adapters are passing these flags.

You can mock functions with tools like Mockery and/or by recreating the functions within the same namespace within which they are invoked in order to determine what arguments are passed. While these take some effort, they also ensure that no breaking changes are introduced later.

Slamdunk:
> Again, in that case, we'd see the addition of the new method during review, giving us time to consider its impact. It's not much different than changing the argument to stream_socket_enable_crypto(), which could also happen via a PR, and which would also spawn a serious review

Maybe I wasn't clear in my example.

1. With a *Trait* and with [Protocol/Smtp.php#L179](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Protocol/Smtp.php#L179) using `$this->getCryptoMethod()` notation you **can** change its behaviour just by adding new coding and you **can't** notice it
1. With a *static call* in [Protocol/Smtp.php#L179](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Protocol/Smtp.php#L179) you **cannot** change its behaviour without changing it and thus noticing it

By the way, I have to admit that I'm a bit confused: stated that I am a TDD-nazi, I feel like you are saying that the strongest and most defensive solution I am proposing isn't good, and that I have to changing it to make it weaker or I have to implement all the tests that this library is missing from the day it was created, and all this for a security issue that usually need to be merged as soon as possible, since PFS and GCM offered in TLSv1.2 is now negated.

I would like to implement all the tests, but I can't in the next few days; likely not before mid june I'll be available for such a huge step.

weierophinney:
@Slamdunk — Regarding this:

> With a Trait and with Protocol/Smtp.php#L179 using $this->getCryptoMethod() notation you can change its behaviour just by adding new coding and you can't notice it

This is simply not true: a pull request would be _introducing a new method in the class_, which _would_ be noticeable. Moreover, it would only affect the _one_ class, not every class composing the trait, which limits the impact to only consumers of that single adapter, should we choose to merge such a patch.

> I feel like you are saying that a stronger and more defensive solution I am proposing isn't good, and that I have to changing it to make it weaker or I have to implement all the tests that this library is missing from the day it was created, and all this for a security issue that usually need to be merged as soon as possible, since PFS and GCM offered in TLSv1.2 is now negated.

We're not saying that the solution you propose is good, or proposing you make it weaker; we're proposing that you modify it to be more consistent with styles and approaches we use throughout the ZF project. What your patch has demonstrated is that _we were missing tests_, which was highlighted by the fact that making the changes you did _did not result in any test breakage_. The proper approach is to add those tests to ensure when a change like this occurs in the future, we can evaluate it properly for backwards compatibility impact.

The approach you suggest has the benefit of being simpler to test. However, if somebody were to modify any of the calls to `stream_socket_enable_crypto()` to no longer call the static method, we'd still have no test failures related to this, which could lead to breakage. That's why we're trying to evaluate whether or not to take a more robust testing approach _now_.

Slamdunk:
I've rewritten the PR to accomplish your wish.

I have to note that I'm against this very implementation, and I've followed what you requested only because it's a right point to have project standards.
Apart from that, I can't consider Traits anything but brittle design code, mainly for this reason https://3v4l.org/3DrmE

About testing `stream_socket_enable_crypto`, the work is big: if it would be up to me, I would wrap `stream_*` functions in a separate Stream class, make it an external dependancy for current protocol classes, and then mock it for protocol classes tests. What would you do?

ezimuel:
@Slamdunk thanks for your work! I think we can definitely improve the unit tests of zend-mail and a possible solution is the one that you suggested. I'll let you know if will be the case. 
Thanks again for your time! I'm going to merge this in a while.

PR #146, opened 2017-05-18
testcase for broken Header\AbstractAddressList::fromString

glensc:
Problem:

1. [GenericHeader](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Header/GenericHeader.php#L33-L40) loads the header in file, decodes it to utf-8
2. the [Headers::get](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Headers.php#L318) attempts to Lazy-Load "To" header class

Lazyload [does stringify and load in from string](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Headers.php#L496-L497)
```php
$encoding = $current->getEncoding();
$headers  = $class::fromString($current->toString());
```
However, toString does not encode comma
AND To header class does split on comma!

see \Zend\Mail\Header\AbstractAddressList::fromString

this PR shows only the problem.

### DIFF
--- a/test/HeadersTest.php
+++ b/test/HeadersTest.php
@@ -401,6 +401,29 @@ class HeadersTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * Problem:
+     * 1. GenericHeader loads the header in file, decodes it to utf-8
+     * 2. the Headers::get attempts to Lazy-Load "To" header class
+     * Lazyload does stringify and load in from string
+     *    $encoding = $current->getEncoding();
+     *    $headers  = $class::fromString($current->toString());
+     * However, toString does not encode comma
+     * AND To header class does split on comma!
+     *
+     * @see \Zend\Mail\Header\AbstractAddressList::fromString
+     */
+    public function testDogFood() {
+        $headers = new Mail\Headers();
+
+        $genericHeader = Mail\Header\GenericHeader::fromString('To: "=?iso-8859-1?Q?W=2C_bj=F8rn?=" <user@example.org>');
+        $this->assertEquals('"W, bjørn" <user@example.org>', $genericHeader->getFieldValue());
+
+        $headers->addHeader($genericHeader);
+        $toHeader = $headers->get('To');
+        $this->assertEquals('To: =?UTF-8?Q?W,=20bj=C3=B8rn?= <user@example.org>', $toHeader->toString());
+    }
+
+    /**
      * @group ZF2015-04
      */
     public function testHeaderCrLfAttackFromString()


glensc:
toString invocation path:
- [GenericHeader::toString](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Header/GenericHeader.php#L171)
- [GenericHeader::getFieldValue](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Header/GenericHeader.php#L144)
- [HeaderWrap::wrap](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Header/HeaderWrap.php#L31)
- [HeaderWrap::wrapUnstructuredHeader](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Header/HeaderWrap.php#L53)
- [HeaderWrap::mimeEncodeValue](https://github.com/zendframework/zend-mail/blob/release-2.7.3/src/Header/HeaderWrap.php#L93)
- [Mime::encodeQuotedPrintableHeader]
- [Mime::_encodeQuotedPrintable]
- ... it got scary here

glensc:
perhaps the fix is in zend-mime project:
https://github.com/zendframework/zend-mime/pull/26

if it gets accepted!

glensc:
after https://github.com/zendframework/zend-mime/pull/26 being applied,
such code with Zend\Mail:
```php
$headerLine = 'To: "=?iso-8859-1?Q?W=2C_bj=F8rn?=" <user@example.org>';
echo Mail\Header\GenericHeader::fromString($headerLine)->toString();
```
before:
```
To: =?UTF-8?Q?"W,=20bj=C3=B8rn"=20<user@example.org>?=
```

now:
```
To: =?UTF-8?Q?"W=2C=20bj=C3=B8rn"=20<user@example.org>?=
```

and
```php
// with headerline without comma:
$headerLine = 'To: =?UTF-8?Q?"W=2C=20bj=C3=B8rn"=20<user@example.org>?=';
echo Mail\Header\To::fromString($headerLine)->toString();
```

```
To: =?UTF-8?Q?"W=2C=20bj=C3=B8rn"?= <user@example.org>
```

glensc:
@weierophinney closed this in zendframework/zend-mime@73e6d05 15 days ago

this is not correct. reopen please.

weierophinney:
@glensc I think I'm not understanding something.

The original string includes an encoded comma (`=2c`).

Why should the comma NOT be encoded when the `To` header is cast to a string? Is it because of the different encodings (iso-8859-1 vs UTF-8)?

(I need to understand why your expectation should be the expected behavior, basically.)

glensc:
@weierophinney it's combination of these two statements from Pull Description:

- However, toString does not encode comma
- AND To header class does split on comma!

so, this seemed the simplest way to solve the problem. it's not forbidden to zealously encode

as https://github.com/zendframework/zend-mime/commit/73e6d05 was accepted in https://github.com/zendframework/zend-mime/pull/26

this just updates unit test data.

test/HeadersTest.php
```diff
@@ -400,6 +400,29 @@ public function testClone()
...
+     *
+     * @see \Zend\Mail\Header\AbstractAddressList::fromString
+     */
+    public function testDogFood() {
+        $headers = new Mail\Headers();
+
+        $genericHeader = Mail\Header\GenericHeader::fromString('To: "=?iso-8859-1?Q?W=2C_bj=F8rn?=" <user@example.org>');
+        $this->assertEquals('"W, bjørn" <user@example.org>', $genericHeader->getFieldValue());
+
+        $headers->addHeader($genericHeader);
```
weierophinney: This test case doesn't make sense given the assertions in your comments and the issue description.

You're making the case that `AbstractAddressList::fromString()` is splitting on a comma, but that `toString()` on such headers is not encoding it.

What I'm seeing here is quite different:

- You're testing the behavior of a `GenericHeader`, not one that is based on `AbstractAddressList`.
- Your assertion is that the expected behavior of `toString()` is NOT to encode the comma.

That's the crux of my confusion: the test is not setting up the conditions you describe, nor testing the expectation you describe.

Can you please clarify?

weierophinney:

michalbundyra:
This repository has been closed and moved to laminas/laminas-mail; a new issue has been opened at https://github.com/laminas/laminas-mail/issues/44.

PR #147, opened 2017-05-20
unable to parse To headers with semicolon separator

glensc:
i wonder why Zend\Mail does not split on semicolon?

- it causes parse error on headers that are semicolon separated: `To: user1@domain.com; user2@domain.org`. to my knowledge outlook sends out such mails: [link](https://blogs.msdn.microsoft.com/oldnewthing/20150119-00/?p=44883)
- it causes "group" feature to be misparsed: `To: undisclosed recipients:; `. [another example](https://github.com/pear/Mail/blob/v1.4.1/Mail/RFC822.php#L62-L64)

this PR adds testcase that should not fail

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All notable changes to this project will be documented in this file, in reverse 
 - [#213](https://github.com/zendframework/zend-mail/pull/213) re-adds support for PHP 5.6 and 7.0; ZF policy is never
   to bump the major version of a PHP requirement unless the package is bumping major version.
 
+### Changed
+
+- Nothing.
+
 ### Deprecated
 
 - Nothing.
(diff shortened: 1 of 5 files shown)


weierophinney:
Your unit test is malformed. Instead of a failing assertion, an exception is being raised due to a line that is longer than allowed.

Additionally, I'm not quite sure how to address this. The `,` character is the one documented as separating fields in the various IETF mail specifications, which makes adding an alternate case harder.

Please try and fix the unit test so that we know we're failing due to the `assertEquals()` assertion, and then we can potentially start working on this.

glensc:
the test case is written with "fixed" code in mind. it will pass if code behaves as expected.

in other words, the code paths used should not throw, but it throws because it encounters different error (field too long)

glensc:
@weierophinney requested changes are made altho i slightly disagree with them!

weierophinney:
Thanks, @glensc, for the test; got a good fix in place now.

PR #148, opened 2017-05-20
AddressList does not parse address with parens in name field properly

glensc:
```
"Supports (E-mail)" <support@example.org>
```

this gets parsed as `Support`, not as expected `Supports (E-mail)`

### DIFF
--- a/src/Address.php
+++ b/src/Address.php
@@ -12,17 +12,56 @@ use Zend\Validator\Hostname;
 
 class Address implements Address\AddressInterface
 {
+    protected $comment;
     protected $email;
     protected $name;
+
+    /**
+     * Create an instance from a string value.
+     *
+     * Parses a string representing a single address. If it is a valid format,
+     * it then creates and returns an instance of itself using the name and
+     * email it has parsed from the value.
+     *
+     * @param string $address
+     * @param null|string $comment Comment associated with the address, if any.
+     * @throws Exception\InvalidArgumentException
+     * @return self
+     */
+    public static function fromString($address, $comment = null)
+    {
+        if (! preg_match('/^((?P<name>.*)<(?P<namedEmail>[^>]+)>|(?P<email>.+))$/', $address, $matches)) {
+            throw new Exception\InvalidArgumentException('Invalid address format');
+        }
+
+        $name = null;
+        if (isset($matches['name'])) {
+            $name = trim($matches['name']);
+        }
+        if (empty($name)) {
+            $name = null;
+        }
+
+        if (isset($matches['namedEmail'])) {
+            $email = $matches['namedEmail'];
+        }
+        if (isset($matches['email'])) {
+            $email = $matches['email'];
+        }
+        $email = trim($email);
+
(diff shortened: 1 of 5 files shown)


glensc:
broken by #12, but it was likely even more broken before that PR.

glensc:
i think `Zend\Mail\Address` should have `$this->comment` property to handle this properly.
currently the `toString` makes information being lost, because it doesn't quote at all!

consider such code:
```php
        $name = "Supports (E-mail)";
        $email = 'support@example.org';
        $address = new \Zend\Mail\Address($email, $name);
        // prints: Supports (E-mail) <support@example.org>
        echo $address->toString(), "\n";

        $to = Header\To::fromString('To:' . $address->toString());
        // prints: To: Supports <support@example.org>
        echo $to->toString(), "\n";
```

weierophinney:
In looking into this, I found this line:

- https://github.com/zendframework/zend-mail/blob/8aac5f7b4d354ea65128be6d2134d4648f47286d/src/Header/AbstractAddressList.php#L62

We're currently _explicitly_ stripping comments from names.

As such, I think you're correct: we should aggregate the comment separately, and then have `toString()` include any comment found. I'll see if I can create a patch.

src/Address.php
```diff
@@ -12,17 +12,56 @@
...
+     */
+    public static function fromString($address, $comment = null)
+    {
+        if (! preg_match('/^((?P<name>.*)<(?P<namedEmail>[^>]+)>|(?P<email>.+))$/', $address, $matches)) {
+            throw new Exception\InvalidArgumentException('Invalid address format');
+        }
+
+        $name = null;
+        if (isset($matches['name'])) {
+            $name = trim($matches['name']);
```
glensc: you could write as: `$name = trim($matches['name']) ?: null` and drop next `if` block ;)

test/AddressListTest.php
```diff
@@ -95,6 +95,18 @@ public function testCanAddManyAddressesAtOnce()
...
+    public function testLosesParensInName()
+    {
+        $header = '"Supports (E-mail)" <support@example.org>';
+
+        $to = Header\To::fromString('To:' . $header);
+        $addressList = $to->getAddressList();
+        $address = $addressList->get('support@example.org');
+        $this->assertEquals('Supports', $address->getName());
+        $this->assertEquals('E-mail', $address->getComment());
+        $this->assertEquals('support@example.org', $address->getEmail());
```
glensc: not testing the `toString`?

src/Address.php
weierophinney: We'll get an error if `$matches['name']` is not set, however. (Tried it previously which is why I went with this.)

test/AddressListTest.php
weierophinney: Probably should! I was primarily ensuring that the various accessors returned what was expected.

src/Address.php
glensc: nah, `$name` is assigned null on line 37 before the `isset` block.

PR #150, opened 2017-05-29
Dummy value in SSL config

jguittard:
Allow defining a dummy value for SSL config in SMTP protocol.
Convenient configuration for continuous delivery workflows (having a generic config file, replacing token values while packaging)


weierophinney:
Thanks, @jguittard

PR #151, opened 2017-05-29
Fixes #149 : correctly trim transport parameters

### LINKED ISSUE #149: Trim error
Line 88 in
 zend-mail/src/Transport/Sendmail.php
must be
`        $string = trim($string);`
instead of
`        trim($string);`

samsonasik:
Fixed #149


weierophinney:
Thanks, @samsonasik

PR #156, opened 2017-06-12
fix regression: MessageId having double brackets

glensc:
this got broken from #86 when header lazyloading was omitted

```
Expected :'<CALTvGe4_oYgf9WsYgauv7qXh2-6=KbPLExmJNG7fCs9B=1nOYg@mail.example.com>'
Actual   :'<<CALTvGe4_oYgf9WsYgauv7qXh2-6=KbPLExmJNG7fCs9B=1nOYg@mail.example.com>>'
```

### DIFF
--- a/test/Storage/MessageTest.php
+++ b/test/Storage/MessageTest.php
@@ -111,6 +111,19 @@ class MessageTest extends TestCase
         $this->assertEquals(substr($message->getContent(), 0, 6), "\t<?php");
     }
 
+    /**
+     * after pull/86 messageId gets double braces
+     *
+     * @see https://github.com/zendframework/zend-mail/pull/86
+     * @see https://github.com/zendframework/zend-mail/pull/156
+     */
+    public function testMessageIdHeader()
+    {
+        $message = new Message(['file' => $this->file]);
+        $messageId = $message->messageId;
+        $this->assertEquals('<CALTvGe4_oYgf9WsYgauv7qXh2-6=KbPLExmJNG7fCs9B=1nOYg@mail.example.com>', $messageId);
+    }
+
     public function testMultipleHeader()
     {
         $raw = file_get_contents($this->file);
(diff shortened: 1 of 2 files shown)


test/Header/MessageIdTest.php
```diff
@@ -14,7 +14,7 @@
 
 /**
  * @group      Zend_Mail
- * @covers Zend\Mail\Header\MessageId<extended>
+ * @covers \Zend\Mail\Header\MessageId<extended>
```
samsonasik: for @covers, already correct without `\` prefix, revert it imo

test/Storage/MessageTest.php
```diff
@@ -20,7 +20,7 @@
 
 /**
  * @group      Zend_Mail
- * @covers Zend\Mail\Storage\Message<extended>
+ * @covers \Zend\Mail\Storage\Message<extended>
```
samsonasik: for @covers, already correct without `\` prefix, revert it imo

test/Header/MessageIdTest.php
asgrim: I believe `@covers` does need `\` prefix

glensc: added it because phpstorm did not complete the class otherwise

test/Storage/MessageTest.php
glensc: added it for ide (phpstorm) to be able to resolve class

test/Header/MessageIdTest.php
samsonasik: @asgrim no, @covers _doesn't_ need it, I use it in my tests without `\`  https://github.com/samsonasik/SanSessionToolbar/blob/master/test/Factory/Controller/SessionToolbarControllerFactoryTest.php#L81-L82 and it detect correctly

Ocramius:
@glensc manually merged into zendframework/zend-mail@d3010b20b3bcc599a8792bb8388845ec154aecf2, thanks!

PR #158, opened 2017-07-19
checking header type interface 

DeyV:
checking header type interface  fix method setType not exists in \ArrayIterator


Ocramius:
This patch requires testing to be considered

weierophinney:
Closing due to lack of response from original author. If you need this patch, please re-submit, with accompanying unit tests.

PR #159, opened 2017-07-20
tests: Mail\Message produced output not usable by Storage\Message

glensc:
this shows the problem zend-mail is having:

mail created with `Mail\Message` and serialized into raw
and then loaded with `Storage\Message` causes parse error if the `$text` contains double unix newlines (`\n\n`).

### DIFF
--- a/test/Storage/MessageTest.php
+++ b/test/Storage/MessageTest.php
@@ -17,6 +17,7 @@ use Zend\Mail\Storage\Exception;
 use Zend\Mail\Storage\Message;
 use Zend\Mime;
 use Zend\Mime\Exception as MimeException;
+use Zend\Mail;
 
 /**
  * @group      Zend_Mail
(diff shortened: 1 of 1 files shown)


test/Storage/MessageTest.php
```diff
@@ -415,6 +416,35 @@ public function testMultivalueToHeader()
...
+     */
+    public function testDogFood() {
+        $text = "Hello, bödi tekst\n\nBye";
+
+        $part = new Mime\Part($text);
+        $part->type = 'text/plain';
+        $part->charset = 'UTF-8';
+
+        $mime = new Mime\Message();
+        $mime->addPart($part);
```
weierophinney: None of the preceding 7 lines are ever used within the test. Is your setup missing something (attaching a MIME part to the message)? Or can the lines be removed safely?

test/Storage/MessageTest.php
```diff
@@ -415,6 +416,35 @@ public function testMultivalueToHeader()
...
+
+        $message = new Mail\Message();
+        $message->setEncoding('UTF-8');
+        $message->setBody($text);
+
+        $raw = $message->toString();
+        $this->assertContains($text, $raw, "original text is preserved");
+
+        $message = new Storage\Message(['raw' => $raw]);
```
weierophinney: I think this is a `Zend\Mime\Decode::splitMessage()` issue.

`Zend\Mail\Storage\Part` calls on that method to split the message into headers and content. My suspicion is that it's searching until the _last_ occurence of the `\n\n` sequence, instead of the _first_.

Can you please try and create and post a reproduce case against the zend-mime component? Thanks in advance!

test/Storage/MessageTest.php
glensc: i think they can be removed as the reproducer worked. otherwise i wouldn't had pushed it this way.

weierophinney:

michalbundyra:
This repository has been closed and moved to laminas/laminas-mail; a new issue has been opened at https://github.com/laminas/laminas-mail/issues/43.

PR #160, opened 2017-07-20
Fixes #73 : allow Message to not has "To" header

### LINKED ISSUE #73: Emails don't need to have a To: header
It's a good idea to add some recipients to an email, but these don't have to be in the To: header. Instead they could be in the "Cc:" or "Bcc:" header lines.

Because of this I think that these checks in Transport/Sendmail.php are not correct:

`if (!$headers->has('to')) {`
`throw new Exception\RuntimeException('Invalid email; contains no "To" header');`
`}`
`$to   = $headers->get('to');`
`$list = $to->getAddressList();`
`if (0 == count($list)) {`
`throw new Exception\RuntimeException('Invalid "To" header; contains no addresses');`
`}`

According to RFC 5322 (http://tools.ietf.org/html/rfc5322) the only absolute necessary (as in MUST) headers are From: and Date:

> The only required header fields are the origination date field and the originator address field(s).  All other header fields are syntactically optional.

Of course there's usually no  email that does not have a recipient, but the check should look for them in To:, Cc: and Bcc:.

samsonasik:
Fixes #73

Check at least one of "To", "Cc", and "Bcc"

### DIFF
--- a/src/Transport/Sendmail.php
+++ b/src/Transport/Sendmail.php
@@ -148,7 +148,8 @@ class Sendmail implements TransportInterface
     {
         $headers = $message->getHeaders();
 
-        if (! ($hasTo = $headers->has('to')) && ! $headers->has('cc') && ! $headers->has('bcc')) {
+        $hasTo = $headers->has('to');
+        if (! $hasTo && ! $headers->has('cc') && ! $headers->has('bcc')) {
             throw new Exception\RuntimeException(
                 'Invalid email; contains no at least one of "To", "Cc", and "Bcc" header'
             );
--- a/src/Transport/Sendmail.php
+++ b/src/Transport/Sendmail.php
@@ -148,8 +148,14 @@ class Sendmail implements TransportInterface
     {
         $headers = $message->getHeaders();
 
-        if (! $headers->has('to')) {
-            throw new Exception\RuntimeException('Invalid email; contains no "To" header');
+        if (! ($hasTo = $headers->has('to')) && ! $headers->has('cc') && ! $headers->has('bcc')) {
+            throw new Exception\RuntimeException(
+                'Invalid email; contains no at least one of "To", "Cc", and "Bcc" header'
+            );
+        }
+
+        if (! $hasTo) {
+            return '';
         }
 
         $to   = $headers->get('to');
--- a/test/Transport/SendmailTest.php
+++ b/test/Transport/SendmailTest.php
@@ -246,15 +246,14 @@ class SendmailTest extends TestCase
         $this->transport->send($message);
     }
 
-    public function testDonotAllowMessageDoesnotHasToAndCcAndBccHeader()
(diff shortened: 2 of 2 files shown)


glensc:
LGTM

oodgaard:
+1

test/Transport/SendmailTest.php
```diff
@@ -223,4 +223,38 @@ public function testTrimmedParameters()
...
+        $message = new Message();
+        $message->addBcc('zf-crteam@lists.zend.com', 'CR-Team, ZF Project')
+                ->setSender('ralph.schindler@zend.com', 'Ralph Schindler')
+                ->setSubject('Testing Zend\Mail\Transport\Sendmail')
+                ->setBody('This is only a test.');
+
+        $this->transport->send($message);
+    }
+
+    public function testDonotAllowMessageDoesnotHasToAndCcAndBccHeader()
```
webimpress: Please fix test method name, I think it should be:
`testDoNotAllowMessageWithoutToAndCcAndBccHeaders`

test/Transport/SendmailTest.php
```diff
@@ -223,4 +223,38 @@ public function testTrimmedParameters()
...
+    {
+        $this->expectException('Zend\Mail\Exception\RuntimeException');
```
webimpress: Please move this before line which should thrown an exception. Like that it's confusing, and the exception could be thrown anywhere below. Can we also use `::class` notation, please?

src/Transport/Sendmail.php
```diff
@@ -148,8 +148,14 @@ protected function prepareRecipients(Mail\Message $message)
     {
         $headers = $message->getHeaders();
 
-        if (! $headers->has('to')) {
-            throw new Exception\RuntimeException('Invalid email; contains no "To" header');
+        if (! ($hasTo = $headers->has('to')) && ! $headers->has('cc') && ! $headers->has('bcc')) {
```
webimpress: I got a bit confused with this assignment in `if` statement, in general I like them, but only when we have single condition, here it's more tricky so I would suggest to move assignment before the `if` statement and then use the variable in the statement with other checks:
```php
$hasTo = $headers->has('to');
if (! $hasTo && ! $headers->has('cc') && ! $headers->has('bcc')) {
```

test/Transport/SendmailTest.php
samsonasik: done

test/Transport/SendmailTest.php
samsonasik: done

src/Transport/Sendmail.php
samsonasik: done

Ocramius:
Manually merged via zendframework/zend-mail@8ca640a946bcc9482af3338482a2f640919a1704, thanks!

PR #163, opened 2017-08-08
Added missing dependency to zend-servicemanager (for --no-dev install)

pce:
If you want to use `zend-mail` as a standalone package and installed it with composer without dev requirements, usage results in a fatal error:  

`PHP Fatal error:  Class 'Zend\ServiceManager\AbstractPluginManager' not found in $project/vendor/zendframework/zend-mail/src/Protocol/SmtpPluginManager.php`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         "zendframework/zend-loader": "^2.5",
         "zendframework/zend-mime": "^2.5",
         "zendframework/zend-stdlib": "^2.7 || ^3.0",
-        "zendframework/zend-validator": "^2.6"
+        "zendframework/zend-validator": "^2.6",
+        "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3"
     },
     "require-dev": {
         "zendframework/zend-config": "^2.6",


froschdesign:
@pce 
It's already there:

```json
"suggest": {
    "ext-intl": "Handle IDN in AddressList hostnames",
    "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3 when using SMTP to deliver messages",
    "zendframework/zend-crypt": "Crammd5 support in SMTP Auth"
},
```

pce:
Could you please explain, how a suggestion solves a unusable package state on a live server?

pce:
@weierophinney i think that is not a user-friendly decision, usually deployment works without suggestions...

froschdesign:
@pce 
> …unusable package state on a live server?
The `zend-mail` component runs without `zend-servicemanager`, if you do not use SMTP to deliver your messages.
If you use SMTP, then you need `zend-servicemanager` and you must install it per composer.

froschdesign:
@pce 
Sorry, but I think you misunderstood the `suggest` part of the composer file. `zend-servicemanger` is not strictly required to use `zend-mail`. Only if you use SMTP.
That's the main idea behind the `suggest` part.
(For a development environment `zend-servicemanger` is needed, because all unit tests must run.)

pce:
In my opinion a "mail" package should include support for SMTP (**the** Mail Transfer Protocol). Can It Be All So Simple :sweat_smile:

glensc:
also you can use smtp without servicemanager if you create the classes instance directly. i do that.

PR #164, opened 2017-08-11
Imap::capability should always return an array

klaussilveira:
If there was a communication failure getting the list of capabilities, the return type should be an `array`, not `false`.


Ocramius:
Please do write a test case

klaussilveira:
I tried. Not really possible without proper mocks and due the way the current suite is setup. It needs a real IMAP server and you can't control it's response from the suite. I only see two possible solutions: abstracting the raw connection logic so read/write methods in this class can be properly tested with mocks OR setting up something like mountebank and mock the socket. 

Any suggestions?

PR #165, opened 2017-08-11
Switch to PHP punycode implementation

mleko:
Switches from depending on ext-intl to true/punycode for the purposes of IDN to ASCII conversion. This change ensures that the functionality will work even when ext-intl is not available.

### DIFF
--- a/src/Header/AbstractAddressList.php
+++ b/src/Header/AbstractAddressList.php
@@ -7,6 +7,7 @@
 
 namespace Zend\Mail\Header;
 
+use TrueBV\Punycode;
 use Zend\Mail\Address;
 use Zend\Mail\AddressList;
 use Zend\Mail\Headers;
@@ -37,6 +38,11 @@ abstract class AbstractAddressList implements HeaderInterface
      * @var string lower case field name
      */
     protected static $type;
+
+    /**
+     * @var Punycode
+     */
+    private static $punycode;
 
     public static function fromString($headerLine)
     {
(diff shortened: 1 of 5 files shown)


glensc:
PR lacks basic information: "why?".

also, the intl dependency was optional, now you added replacement dependency as required.

you should also post what new dependencies got installed with this PR to help maintainers decide fate of this PR.

mleko:
Composer dependency is easier to install than PECL extension. Also it don't require any special privileges.

glensc:
don't "answer" it, put it to PR description!

src/Header/AbstractAddressList.php
```diff
@@ -96,15 +102,19 @@ public function getFieldName()
...
     protected function idnToAscii($domainName)
     {
-        if (extension_loaded('intl')) {
-            return (idn_to_ascii($domainName) ?: $domainName);
+        if (null === self::$punycode) {
+            self::$punycode = new Punycode();
+        }
+        try {
+            return self::$punycode->encode($domainName);
+        } catch (\Exception $e) {
```
Ocramius: Is there no more specific exception type?

src/Header/AbstractAddressList.php
```diff
@@ -96,15 +102,19 @@ public function getFieldName()
...
-     * @param string $domainName  the UTF-8 encoded email
+     * @param string $domainName the UTF-8 encoded email
      * @return string
      */
     protected function idnToAscii($domainName)
     {
-        if (extension_loaded('intl')) {
-            return (idn_to_ascii($domainName) ?: $domainName);
+        if (null === self::$punycode) {
+            self::$punycode = new Punycode();
```
Ocramius: `if (! self::$punycode) {`

src/Header/AbstractAddressList.php
```diff
@@ -39,6 +40,11 @@
      */
     protected static $type;
 
+    /**
+     * @var Punycode
```
Ocramius: `|null`

mleko: Good point, thank you

src/Header/AbstractAddressList.php
mleko: I'm not sure if it's beneficial but I found one

src/Header/AbstractAddressList.php
mleko: I prefer my version. Is there style guide somewhere?

weierophinney:
Thanks, @mleko!

PR #171, opened 2017-09-15
Add PHPDoc comments

danielabyan:
In some cases, the IDE can not recognize the types of variables. To solve this problem, I've added comments in some places.
And I added comments for the following methods
 - Add missing comment for method GenericHeader::fromString
 - Add PHPDoc @return comment for Smtp::setUseCompleteQuit() method
 - PHPDoc @return comment does not apply to the constructor because remove this comment for Address::__construct
 - Add Mime\Message type for Mail\Message::$body property.

### DIFF
--- a/src/Address.php
+++ b/src/Address.php
@@ -23,7 +23,6 @@ class Address implements Address\AddressInterface
      * @param  string $email
      * @param  null|string $name
      * @throws Exception\InvalidArgumentException
-     * @return Address
      */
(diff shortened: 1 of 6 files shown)


Ocramius:
@danielabyan thanks! Manually merged to zendframework/zend-mail@0c37b5983730e0d25cfd2a7712519050e4fef6e9

PR #172, opened 2017-10-17
Add a time limit option for the SMTP connection

MatthiasKuehneEllerhold:
As described in the documentation, this PR will add the option for a time limit for the SMTP connection.

If you like the general direction, I'm willing to add unit tests tomorrow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Added
 
+- [#172](https://github.com/zendframework/zend-mail/pull/172) adds the flag `connection_time_limit` to the possible `Zend\Mail\Transport\Smtp` options.
+  This flag, when provided as a positive integer, and in conjunction with the `use_complete_quit` flag, will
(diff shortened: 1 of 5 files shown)


Ocramius:
This patch requires extended testing (probably via a `clock` object)

MatthiasKuehneEllerhold:
I've added a test with the help of Reflections.

There may be one issue: this PR sets the current connection to `null` if the time limit was reached. This means we're relying on `lazyLoadConnection()` to recreate the connection. But we cant recreate the connection if it was injected through `setConnection()`. So (maybe / probably) we're getting a different connection if the time limit is up and the connection is recreated using `lazyLoadConnection()` if it was injected through `setConnection()` on the first mail.

Example: We're injected a mocked connection in the Unit-Tests and do not rely on `lazyLoadConnection()`.

Possible solution: Implement the auto-reconnect in the `Zend\Mail\Protocol\Smtp` class instead.

weierophinney:
Thanks, @MatthiasKuehneEllerhold!

PR #179, opened 2017-11-02
Header Parsing Error

rodnaph:
I've added a failing test case for an error I'm running into in strict mode with real email data, not sure on the options for fixing if it's an invalid header? Or suppressing the error would be ok? Or more robust parsing of the header is needed?

The problem is the semi-colon being used as a delimiter for splitting, when it seems it can be valid to occur in pair values when quoted as a literal...

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -23,6 +23,7 @@
 
     <php>
         <ini name="date.timezone" value="UTC"/>
+        <ini name="error_reporting" value="-1"/>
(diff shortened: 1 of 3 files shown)


rodnaph:
Seems the value of these properties can be a quoted string: https://tools.ietf.org/html/rfc1341

rodnaph:
Superseded by https://github.com/zendframework/zend-mail/pull/211

PR #180, opened 2017-11-10
added missing new keyword

uikolas:
As far I understand, these are not functions (maybe i'm not correct?), but classes, so added missing new keyword


Ocramius:
@uikolas thanks! Manually merged in zendframework/zend-mail@7fe57dceb0c07599c30c181cd8e1a96687ef9e01

Issue #187, opened 2017-12-28
header lazy loading breaks header access


glensc:
take this example code, after calling `$message->getHeaders()` to lazy-load the headers, the `DKIM-Signature` header can not be retrieved:

```php
<?php

require __DIR__.'/vendor/autoload.php';
use PHPUnit\Framework\TestCase;
use Zend\Mail\Storage;

$headerValue = "v=1; a=rsa-sha25; c=relaxed/simple; d=example.org; h=\r\n\tcontent-language:content-type:content-type:in-reply-to";
$headers = "DKIM-Signature: {$headerValue}";

$message = new Storage\Message(['headers' => $headers, 'content' => 'irrelevant']);
$h = $message->getHeaders();
// calling toString will lazy load all headers
// and break DKIM-Signature
$h->toString();
$x = $h->get('DKIM-Signature');
// not reached here
```

i've narrowed it down to [iconv_mime_decode](http://php.net/iconv_mime_decode) in [Zend\Mail\Header\HeaderWrap::mimeDecodeValue](https://github.com/zendframework/zend-mail/blob/release-2.8.0/src/Header/HeaderWrap.php#L106)

seems it strips `\n` for no apparent reason, leaving `\r` in place, thus breaking header continuation sequence (`\r\n\t`) (`less` is used to visually see `\r`):

```
$ php71 -r 'var_dump(iconv_mime_decode($s="v=1; a=rsa-sha256; c=relaxed/simple; d=example.net; h=\r\n bb", ICONV_MIME_DECODE_CONTINUE_ON_ERROR, "UTF-8"));var_dump($s);'|less

string(58) "v=1; a=rsa-sha256; c=relaxed/simple; d=example.net; h=^M bb"
string(59) "v=1; a=rsa-sha256; c=relaxed/simple; d=example.net; h=
 bb"
```

glensc:
exception trace from `mimeDecodeValue` method (maybe useful):

```
#0 zend-mail/src/Header/GenericHeader.php(36): Zend\Mail\Header\HeaderWrap::mimeDecodeValue('v=1; a=rsa-sha2...')
#1 zend-mail/src/Headers.php(482): Zend\Mail\Header\GenericHeader::fromString('DKIM-Signature:...')
#2 zend-mail/src/Headers.php(231): Zend\Mail\Headers->loadHeader('DKIM-Signature:...')
[...]

PR #188, opened 2017-12-28
unfold before calling iconv_mime_decode

glensc:
proposed solution for #187

resolves: #187


Ocramius:
@glensc thanks! And thanks for the streak of impeccable patches (honestly: very well researched and tested patches!)

Manually merged via zendframework/zend-mail@452735048ff41d177bdac89a117edf50e6e27797

PR #190, opened 2018-01-05
PHPDoc: `Message#getBody()` can also return a `string`

Slamdunk:
Body can also be a string:
https://github.com/zendframework/zend-mail/blob/d787bf7e2ad88bdd79fbf5431b35acbac79408a0/src/Message.php#L17-L22

### DIFF
--- a/src/Message.php
+++ b/src/Message.php
@@ -415,7 +415,7 @@ class Message
     /**
      * Return the currently set message body
      *
-     * @return object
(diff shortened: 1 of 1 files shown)


Ocramius:
@Slamdunk thanks! Manually merged via zendframework/zend-mail@0b7b501b1ba0bcbcb6edb1a31cac26bb31d5a19e

PR #189, opened 2017-12-29
Add PHP 7.2 support

Slamdunk:
- [x] Adopt recent `.travis.yml` ZF standards ([zend-session/.travis.yml](https://github.com/zendframework/zend-session/blob/develop/.travis.yml) picked)
- [x] Test PHP 7.2
- [x] Add PHP 7.2 support
    1. `idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated`

Fix https://github.com/zendframework/zend-mail/issues/177, https://github.com/zendframework/zend-mail/issues/183

Supersedes https://github.com/zendframework/zend-mail/pull/181, https://github.com/zendframework/zend-mail/pull/192

PR #181, opened 2017-11-18
Fix deprecated variant in idn_to_ascii

matbech:
In PHP 7.2, INTL_IDNA_VARIANT_2003 is deprecated in 
PHP Deprecated:  idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated in zendframework\zend-mail\src\Header\AbstractAddressList.php on line 105

Reference issue: https://github.com/zendframework/zend-mail/issues/177

PR #192, opened 2018-01-22
Use the INTL_IDNA_VARIANT_UTS46 idn_to_ascii variant

mtorromeo:
The default variant INTL_IDNA_VARIANT_2003 is deprecated in php 7.2.
The default will be changed to INTL_IDNA_VARIANT_UTS46 in php 7.4 so we should use that if it is defined.
This is similar to how it was changed in `Zend\Validator\EmailAddress`.


Slamdunk:
Duplicate of #181 and #189

Ocramius:
Handled in #189

PR #196, opened 2018-03-01
Fix #92 : Allow whitespace empty in empty single line header

### LINKED ISSUE #92: "Malformed header detected" on valid example from rfc7103 7.3
Per [RFC 7103 section 7.3](https://tools.ietf.org/html/rfc7103#section-7.3), having whitespace on an otherwise empty line is deprecated but still valid.  I'm running into this with emails generated by some Microsoft product. 

Code to reproduce the error:

``` php
require __DIR__.'/vendor/autoload.php';

$src = "From: user@example.com
To: userpal@example.net
Subject: This is your reminder

  about the football game tonight
Date: Wed, 20 Oct 2010 20:53:35 -0400

Don't forget to meet us for the tailgate party!
";
new Zend\Mail\Storage\Message(array('raw' => $src));
```

This results in:

```
PHP Fatal error:  Uncaught exception 'Zend\Mail\Exception\RuntimeException' with message 'Malformed header detected' in vendor/zendframework/zend-mail/src/Headers.php:88
vendor/zendframework/zend-mail/src/Headers.php:88
vendor/zendframework/zend-mime/src/Decode.php:142
vendor/zendframework/zend-mail/src/Storage/Part.php:105
vendor/zendframework/zend-mail/src/Storage/Message.php:54
```

 One workaround is to use pipe the message source through formail first and tell it to merge header lines:  
`formail -c -f`

samsonasik:
Fix #92

### DIFF
--- a/test/Storage/MessageTest.php
+++ b/test/Storage/MessageTest.php
@@ -145,7 +145,7 @@ class MessageTest extends TestCase
     public function testAllowWhitespaceInEmptySingleLineHeader()
     {
         $src = "From: user@example.com\nTo: userpal@example.net\nSubject: This is your reminder\n  \n  about the football game tonight\nDate: Wed, 20 Oct 2010 20:53:35 -0400\n\nDon't forget to meet us for the tailgate party!\n";
-        $message = new Message(array('raw' => $src));
+        $message = new Message(['raw' => $src]);
 
         $this->assertEquals(
             'This is your reminder about the football game tonight',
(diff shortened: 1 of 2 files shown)


chrisdeeming:
Unfortunately this isn't the totally correct fix as far as I can tell. It will work in some cases, but only if there are no more than 1 empty lines. We're seeing examples of headers that contain upwards of 5-10 empty lines.

Admittedly, that kind of email is clearly in the wrong in terms of the spec (though it seems a bit of a grey area in places) but there are libraries out there which seem to be able to parse such emails correctly without falling over on an arbitrary limit of empty lines.

test/Storage/MessageTest.php
```diff
@@ -141,6 +142,25 @@ public function testMultipleHeader()
         );
     }
 
+    public function testAllowWhitespaceInEmptySingleLineHeader()
+    {
+        $src = "From: user@example.com\nTo: userpal@example.net\nSubject: This is your reminder\n  \n  about the football game tonight\nDate: Wed, 20 Oct 2010 20:53:35 -0400\n\nDon't forget to meet us for the tailgate party!\n";
+        $message = new Message(array('raw' => $src));
```
glensc: consistency: short array syntax

samsonasik: done ;)

samsonasik:
@chrisdeeming that probably an edge case that should not be handled ^^

weierophinney:
Thanks, @samsonasik.

PR #197, opened 2018-03-02
Allow zend-crypt 3 and phpunit 7

remicollet:


composer.json
```diff
@@ -23,8 +23,8 @@
     "require-dev": {
         "zendframework/zend-config": "^2.6",
         "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3",
-        "zendframework/zend-crypt": "^2.6",
-        "phpunit/phpunit": "^6.0.8 || ^5.7.15",
+        "zendframework/zend-crypt": "^2.6 || ^3.0",
+        "phpunit/phpunit": "^7.0 || ^6.0.8 || ^5.7.15",
```
webimpress: This is dev dependency and this is a testing framework, we can use here just the latest version - `^7.1.4` as that repository supports only PHP 7.1+.

webimpress:
We need also regenerate composer.lock file because now we are getting:
```
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them.
```

(run `composer update nothing` and commit `composer.lock` file).

glensc:
the command is actually `composer update --lock`

but aside, i don't think libraries should track composer.lock
https://getcomposer.org/doc/02-libraries.md#lock-file

weierophinney:
@glensc —

> but aside, i don't think libraries should track composer.lock

We feel otherwise, as tracking it has allowed us to identify when:

- we use library features that are not available in the lowest version of the library we require (generally difference between locked and lowest).
- when a newer version of a library has a BC break that affects our usage of it (difference between locked and latest).

We have plenty of experience with both scenarios. In point of fact: due to the fact that we use this approach, I was able, just yesterday, to isolate test failures to a behavior change in PHPUnit. I would not have found this change as easily without the testing strategy we use.

PR #199, opened 2018-03-13
Protocol::setUseCompleteQuit PHPDoc minor fix

Slamdunk:


samsonasik:
`dev-master` already use php:^7.1 as requirement, so probably it is better to change all codebase with type hinting

webimpress:
@samsonasik adding typehints will be BC break, so we can do it only for next major release.

weierophinney:
Thanks, @Slamdunk.

PR #204, opened 2018-04-19
Handle multiline multibyte headers correctly

### LINKED ISSUE #194: HeaderWrap::mimeDecodeValue unable to handle multiline utf header splitted across character
I've encountered email with subject, that was not handled by *zend-mail* properly. After some investigation I found that header value was split in middle of **multibyte** utf character. 

Test to reproduce it:

```php
    public function testMultilineWithMultibyteSplitAcrossCharacter()
    {
        $originalValue = 'аф';

        $this->assertEquals(strlen($originalValue), 4);

        $part1 = base64_encode(substr($originalValue, 0, 3));
        $part2 = base64_encode(substr($originalValue, 3));

        $header = 'Subject: =?utf-8?B?' . $part1 . '?==?utf-8?B?' . $part2 . '?=';

        $this->assertEquals(
            $originalValue,
            GenericHeader::fromString($header)->getFieldValue()
        );
    }
```
I fix it locally:

```php
    public static function mimeDecodeValue($value)
    {
        return array_reduce(
            imap_mime_header_decode(imap_utf8($value)),
            function ($acc, $item) { return $acc . $item->text; },
            ''
        );
    }
```

If depending on *imap* extension is ok, I will make PR. Or we can check if *imap* extension is enabled, otherwise use `iconv_mime_decode` as before.

fred84:
Fixes #194

### DIFF
--- a/src/Header/HeaderWrap.php
+++ b/src/Header/HeaderWrap.php
@@ -114,8 +114,8 @@ abstract class HeaderWrap
         if (self::isNotDecoded($value, $decodedValue) && extension_loaded('imap')) {
             return array_reduce(
                 imap_mime_header_decode(imap_utf8($value)),
-                function ($acc, $item) {
-                    return $acc . $item->text;
+                function ($accumulator, $headerPart) {
+                    return $accumulator . $headerPart->text;
                 },
                 ''
             );
--- a/src/Header/HeaderWrap.php
+++ b/src/Header/HeaderWrap.php
@@ -126,9 +126,9 @@ abstract class HeaderWrap
 
     private static function isNotDecoded($originalValue, $value)
     {
-        return substr($value, 0, 2) == '=?'
-            && substr($value, -2, 2) == '?='
-            && strpos($originalValue, $value) !== false;
+        return 0 === strpos($value, '=?')
+            && strlen($value) - 2 === strpos($value, '?=')
+            && false !== strpos($originalValue, $value);
     }
 
     /**
--- a/src/Header/HeaderWrap.php
+++ b/src/Header/HeaderWrap.php
@@ -124,12 +124,13 @@ abstract class HeaderWrap
         return $decodedValue;
     }
 
-    private static function isNotDecoded($originalValue, $value) {
+    private static function isNotDecoded($originalValue, $value)
+    {
         return substr($value, 0, 2) == '=?'
             && substr($value, -2, 2) == '?='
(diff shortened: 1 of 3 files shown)


test/Header/HeaderWrapTest.php
```diff
@@ -111,4 +111,25 @@ public function testCanBeEncoded()
         $res = HeaderWrap::canBeEncoded($value);
         $this->assertTrue($res);
     }
+
+    public function testMultilineWithMultibyteSplitAcrossCharacter()
+    {
+        if (! extension_loaded('imap')) {
+            $this->markTestSkipped('works only with imap extension enabled');
+        }
```
glensc: perhaps phpunit annotation instead?

```
@requires extension imap
```

fred84: @glensc Thanks for suggestion! PR updated.

src/Header/HeaderWrap.php
```diff
@@ -112,9 +112,27 @@ public static function mimeDecodeValue($value)
 
         $decodedValue = iconv_mime_decode($value, ICONV_MIME_DECODE_CONTINUE_ON_ERROR, 'UTF-8');
 
+        // imap (unlike iconv) can handle multibyte headers which are splitted across multiple line
+        if (self::isNotDecoded($value, $decodedValue) && extension_loaded('imap')) {
+            return array_reduce(
+                imap_mime_header_decode(imap_utf8($value)),
+                function ($acc, $item) {
```
glensc: perhaps more useful variable names than `$acc` and `$item`?

fred84: updated

src/Header/HeaderWrap.php
```diff
@@ -112,9 +112,27 @@ public static function mimeDecodeValue($value)
...
+                ''
+            );
+        }
+
         return $decodedValue;
     }
 
+    private static function isNotDecoded($originalValue, $value)
+    {
+        return substr($value, 0, 2) == '=?'
```
samsonasik: can use `0 === strpos($value, '=?')`

fred84: done

fred84:
@samsonasik @glensc Some tests on DEPS=lowest are broken, but it's unrelated to my changes. Master branch also broken.

weierophinney:
Thanks, @fred84!

PR #211, opened 2018-05-16
Content Type Header Parsing

rodnaph:
When parsing the `Content-Type` header there is an assumption that the type and key/value pairs can be split by semi-colon, but this is incorrect when a value contains a semi-colon (eg. `name="foo; bar"`). This format is covered by the RFC (https://tools.ietf.org/html/rfc2045#page-12, `Must be in quoted-string to use within parameter values`), and is seen in data in the wild.  This patch adds support for this.

Currently an exception is thrown in strict mode because of an undefined offset (`Undefined offset: 1`). I've updated the code to use `parse_str` to better handle this data, which I think is an improvement over the regexp splitting, but still not ideal without a proper lexer/tokenizer obv.

I've also enabled strict mode for PHPUnit by default to expose these kinds of bugs in the test suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
+- [#211](https://github.com/zendframework/zend-mail/pull/211) fixes how the `ContentType` header class parses the value it receives. Previously,
+  it was incorrectly splitting the value on semi-colons that were inside quotes; in now correctly
+  ignores them.
+
 - [#204](https://github.com/zendframework/zend-mail/pull/204) fixes `HeaderWrap::mimeDecodeValue()` behavior when handling a multiline UTF-8
   header split across a character. The fix will only work when ext-imap is present, however.
 
--- a/src/Header/AddressListParser.php
+++ b/src/Header/AddressListParser.php
@@ -17,9 +17,10 @@ class AddressListParser
 
     /**
      * @param string $value
+     * @param array $delims
      * @return array
      */
-    public static function parse($value)
+    public static function parse($value, $delims = self::CHAR_DELIMS)
     {
         $values            = [];
         $length            = strlen($value);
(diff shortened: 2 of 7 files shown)


bendavies:
@weierophinney any chance to take a look? thanks!

weierophinney:
@rodnaph Please rebase off of current master, and then look at https://github.com/zendframework/zend-mail/blob/79f0f72fbaf606f4a2db322bb2fe2e381d2e9d53/src/Header/AddressListParser.php — this class provides a parser for address list items currently, and I think the approach it presents (a proper parser) is likely a better fit for the fix you're trying to implement.

(Also, run `composer cs-check` and/or `composer cs-fix` before pushing again, to get rid of CS issues.)

rodnaph:
@weierophinney Thanks - the parser class works for this case if you can allow a minor customisation (though the name is now perhaps a little misleading if it's used in both locations).

src/Header/AddressListParser.php
```diff
@@ -17,9 +17,10 @@ class AddressListParser
 
     /**
      * @param string $value
+     * @param array $delims
      * @return array
      */
-    public static function parse($value)
+    public static function parse($value, $delims = self::CHAR_DELIMS)
```
bendavies: array type hint?

test/Header/ContentTypeTest.php
```diff
@@ -45,6 +45,29 @@ public function testExtractsExtraInformationWithoutBeingConfusedByTrailingSemico
         $this->assertEquals($header->getParameters(), ['name' => 'foo.pdf']);
     }
 
+    public static function getLiteralData()
+    {
+        return [
```
bendavies: need long array syntax for php 5.6 compat

weierophinney:
Since the `AddressListParser` has not yet been released, I've taken the liberty of renaming it to `ListParser`, and pushing the changes back to your branch. With those in place, I think this is ready to go.

CHANGELOG.md
```diff
@@ -42,6 +42,10 @@ All notable changes to this project will be documented in this file, in reverse
 
 ### Fixed
 
+- [#211](https://github.com/zendframework/zend-mail/pull/211) fixes how the `ContentType` header class parses the value it receives. Previously,
+  it was incorrectly splitting the value on semi-colons that were inside quotes; in now correctly
```
bendavies: *it* now correctly ignores them.

PR #224, opened 2019-03-04
Add test for #147 regression

glensc:
Add test for #147 regression:

Looks like the new parser class treats `"` and `'` equally, thus `'"Bob O\'Reilly" <bob@example.com>,blah@example.com'` does not parse properly.

cc @weierophinney


glensc:
will be merged with https://github.com/zendframework/zend-mail/pull/226 thanks @silvadiego

PR #227, opened 2019-06-13
 add noOverrideParameters method not to override parameter

tzmfreedom:
`Zend\Mail\Transport\SendMail` always append envelope-from parameter.  
But I want to set any envelope-from.  
In this PR, we will can ignore append feature and use the raw parameter.

---

Provide a narrative description of what you are trying to accomplish:

- [x] Are you creating a new feature?
  - [x] Why is the new feature needed? What purpose does it serve?
  - [x] How will users use the new feature?
  - [x] Base your feature on the `develop` branch, and submit against that branch.
  - [x] Add only one feature per pull request; split multiple features over multiple pull requests
  - [x] Add tests for the new feature.
  - [ ] Add documentation for the new feature.
  - [ ] Add a `CHANGELOG.md` entry for the new feature.


glensc:
without getting deeper to the actual changes, have some thoughts:

1. seems weird to have a negative parameter. how about set positive parameter which is b default true, and can be set as false. 
2. maybe have a specific option (envelope related) rather generic "no override" as behavior could be different if want/want not to override some other parameter.

tzmfreedom:
Thank you for you reply and advice !

I rethink, revise my code and changed this PR goal. 

On this PR, only when not set option `-f`, set `-f` option automatically.  
This means if I set `-ffoo@example.com` or `-bs -ffoo@example.com`, Transport\Sendmail doesn't set `-f` option.  
So there is no negative parameter, no additional method.

glensc:
you should update PR title as well. and the body is a bit hard to read as well.

weierophinney:
This repository has been closed and moved to laminas/laminas-mail; a new issue has been opened at https://github.com/laminas/laminas-mail/issues/17.

weierophinney:

PR #230, opened 2019-07-06
Fix AddressList toString method to quote semicolon

glensc:
Certain input of `AddressList` headers, cannot be converted to string and back to header object because of incorrect quoting:

```
From: "Foo;" <root@example.com>
```

gets incorrectly converted as
```
From: Foo; <root@example.com>
```

but `;` is address separator, so it needs to be quoted:

```
From: "Foo;" <root@example.com>
```

The problem I discovered internally when using `Storage\Message` with `Headers` input, therefore testing that method is included in the unit test:

```php
$message = new Message(['headers' => new Headers(), 'content' => (string)$body]);

// Mime\Decode::splitMessage calls toString on headers object which creates invalid result:
if ($message instanceof Headers) {
    $message = $message->toString();
}
```


glensc:
@weierophinney can you take care of this?

glensc:
this is a regression from #147 as it added `;` separator support that's why (i believe) this exception is caught.

however, there should not be an assumption that only zend-mail is used for reading writing mail messages.

glensc:
or @Xerkus

weierophinney:
This repository has been closed and moved to laminas/laminas-mail; a new issue has been opened at https://github.com/laminas/laminas-mail/issues/14.

weierophinney:

PR #235, opened 2019-08-01
Updates configuration of MkDocs for Documentation

froschdesign:


mkdocs.yml
```diff
@@ -17,6 +17,5 @@ pages:
             - "File Transport Options": transport/file-options.md
         - "Reading and Storing Mail": read.md
 site_name: zend-mail
-site_description: zend-mail
+site_description: 'Parse, create, store, and send email messages, using a variety of storage and transport protocols.'
```
webimpress: @froschdesign are we fine with using single quotes here? I've seen usually we go with double

froschdesign: Both are working: https://pyyaml.org/wiki/PyYAMLDocumentation#scalars

The single quotes are correct here because no escaping is needed. But we were never very strict in this topic.

webimpress:
Thanks, @froschdesign!

Issue #84, opened 2016-04-25
[SSL/TLS] Support for pass the stream context


gongo:
From PHP 5.6, SSL/TLS client streams now enable peer verification by default.
So, error occurs when communicate with the server that is using a self-signed certificate.

See: http://php.net/manual/migration56.openssl.php

In order to avoid the above problem:

```php
$options = [
    'ssl' => [
        'verify_peer_name' => false,
        'verify_peer'      => false,
    ]
];

$context = stream_context_create($options);

// ex. https://github.com/zendframework/zend-mail/blob/release-2.7.0/src/Protocol/Pop3.php#L95

// $this->socketfsockopen($host, $port, $errno, $errstr, self::TIMEOUT_CONNECTION);
$this->socketfsockopen($host, $port, $errno, $errstr, self::TIMEOUT_CONNECTION, $context);
```

In this way, there is a need to pass a [stream context](http://php.net/stream_context_create) to the stream function.

However, the current `Zend\Mail\Protocol\Pop3` and `Zend\Mail\Protocol\Smtp` look like does not provide an interface to pass stream context.
Is there support for the above interface in future?

---

A similar way: [zend-http/zend.http.client.adapters.md at release-2.5.4 · zendframework/zend-http](https://github.com/zendframework/zend-http/blob/release-2.5.4/doc/book/zend.http.client.adapters.md#customizing-and-accessing-the-socket-adapter-stream-context)

Cruiser13:
I created a fork which solves this issue hard-coded here: https://github.com/Cruiser13/zend-mail
I don't think they'll add a config solution inside the core since ZF1 reached end of lifetime.

weierophinney:
Um... This component is v2, and still maintained...

Cruiser13:
Sorry, I meant to post it here, got the versions mixed up: https://github.com/zendframework/zf1/issues/709
So there's hope for ZF2 and this issue :)

jensstalder:
Same here. Would be great if we could disable verify_peer for `Zend\Mail\Protocol\Pop3` `Zend\Mail\Protocol\Smtp` `Zend\Mail\Protocol\Imap`. This issue appears to be open for 2 years?

michalbundyra:
This repository has been closed and moved to laminas/laminas-mail; a new issue has been opened at https://github.com/laminas/laminas-mail/issues/63.

PR #246, opened 2019-10-18
Add option novalidatecert to connect()

abulhol:
This commit is a proposed fix for the following issue:
https://github.com/zendframework/zend-mail/issues/84
It introduces an optional parameter `novalidatecert` both for POP3 and IMAP, which makes it possible to  connect to mail servers with self-signed certificates, e.g. for development purposes.

### DIFF
--- a/src/Protocol/Imap.php
+++ b/src/Protocol/Imap.php
@@ -62,7 +62,7 @@ class Imap
      * @throws Exception\RuntimeException
      * @return string welcome message
      */
-    public function connect($host, $port = null, $ssl = false)
+    public function connect($host, $port = null, $ssl = false, $novalidatecert = false)
     {
         $isTls = false;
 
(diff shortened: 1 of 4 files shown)


webimpress:
Change of method signature is BC Break. Can we implement it the other way without changing the method signature?

abulhol:
> Change of method signature is BC Break. Can we implement it the other way without changing the method signature?

This is why I added this as an optional parameter, which is indeed BC because the params are an array. 
How else is the method caller supposed to signal that SSL certificates should not be checked, if not through the method's arguments?

abulhol:
I see, thanks! You can close this PR, I will try to implement it like this and create a new PR.

webimpress:
@abulhol no need for new PR, you can continue on this.
Please also remember to update the documentation !

Thanks again :)

webimpress:
@abulhol Also, remember to target `develop` branch instead of `master` as it is feature, so will be planned for next minor release (you can change it on editing PR, but you need also rebase the branch).

abulhol:
I have created a PR to the `develop` branch (I did not manage to redirect this one here).
Where do I update the documentation? I mean this here: https://docs.zendframework.com/zend-mail/

froschdesign:
@abulhol 
See the Markdown files in https://github.com/zendframework/zend-mail/tree/master/docs/book

PR #247, opened 2019-10-18
Add option novalidatecert to connect()

abulhol:
See discussion on PR #246

### DIFF
--- a/src/Protocol/Imap.php
+++ b/src/Protocol/Imap.php
@@ -62,7 +62,7 @@ class Imap
      * @throws Exception\RuntimeException
      * @return string welcome message
      */
-    public function connect($host, $port = null, $ssl = false)
+    public function connect($host, $port = null, $ssl = false, $novalidatecert = false)
     {
         $isTls = false;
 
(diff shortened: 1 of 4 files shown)


src/Protocol/Imap.php
```diff
@@ -18,6 +18,12 @@ class Imap
      */
     const TIMEOUT_CONNECTION = 30;
 
+    /**
+     * Do not validate the SSL certificate if set to true
+     * @var null|bool
+     */
+    public $novalidatecert;
```
froschdesign: As long as we do not have typed properties, we can not use public here. You can use the constructor to set a value for this property.

weierophinney: Agreed. I'd add either:

- a 4th constructor argument, `$validateCert = true`
- a 4th constructor argument, `array $options`, and document a "validate_cert" or "validateCert" option.

The latter makes it easier to expand options, but leads to more validation once we add typehints. The former is self-documenting, but means if we add more options down the line, the constructor signature gets really large.

I'm fine with either approach, but agree with @froschdesign here that a public property is not a great idea.

abulhol: Hi guys, thank you for the feedback! 
I have added the variable to the constructor, but this does not really fix the issue. The thing is that the way you create a POP3 or IMAP connection is by creating an instance of `Zend\Mail\Storage\Pop3` or `Zend\Mail\Storage\Imap3`. 
The constructor then calls the Zend\Mail\Protocol constructor without arguments, followed by calling `connect()`. 
But because I cannot add a new parameter to `connect()` (see previous PR), and I have to leave the constructor call without arguments as it is, I have now added a setter method for `novalidatecert`. 
Please see my latest commit.

abulhol:
Any progress or feedback on this PR?

webimpress:
@abulhol 

> Any progress or feedback on this PR?

As you can see there is no any new comments or approvals, so not yet :)

weierophinney:
This repository has been closed and moved to laminas/laminas-mail; a new issue has been opened at https://github.com/laminas/laminas-mail/issues/5.

weierophinney:

